PR #4, opened 2019-05-30
Allow building eBPF from the main builder image

albertvaka:
Brings dependencies from the future into our builder images.

Kernel headers come from Debian 9 for the deb builder and Fedora 26 for rpm.

It's a bit of a hack.

### DIFF
--- a/deb-x64/Dockerfile
+++ b/deb-x64/Dockerfile
@@ -103,6 +103,11 @@ RUN curl -O http://releases.llvm.org/${CLANG_VERSION}/clang+llvm-${CLANG_VERSION
     tar -xf clang+llvm-${CLANG_VERSION}-x86_64-linux-sles11.3.tar.xz --no-same-owner --strip 1 -kC /usr/ && \
     rm clang+llvm-${CLANG_VERSION}-x86_64-linux-sles11.3.tar.xz
 
+# To build the EBPF code we need kernel headers for Linux 4.9
+RUN curl -Sl -O https://dd-agent-omnibus.s3.amazonaws.com/kernel-4.9-headers-deb-x64.tgz && \
+    tar xf kernel-4.9-headers-deb-x64.tgz --no-same-owner --strip 1 -C /usr && \
+    rm kernel-4.9-headers-deb-x64.tgz
+
 # Entrypoint
 COPY ./entrypoint.sh /
 RUN chmod +x /entrypoint.sh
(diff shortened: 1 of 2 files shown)


albertvaka:
This same patch applies to the ARM builder images.

I think getting an artifact from S3 is more cumbersome as we expand the archs we support (eg: in case we are building the agent for ARM we need to be able to download a system-probe built for ARM as well), plus it makes it harder for customers who want to build the agent from sources to do so.

kiyose:
This is blocking building the 6.12.1 agent from source.

deb-x64/Dockerfile
```diff
@@ -95,6 +95,22 @@ RUN set -ex \
...
+# Version built for sles11 is linked against an old enough glibc to be used from this image
+RUN curl -O http://releases.llvm.org/8.0.0/clang+llvm-8.0.0-x86_64-linux-sles11.3.tar.xz && \
+    tar -xf clang+llvm-8.0.0-x86_64-linux-sles11.3.tar.xz --no-same-owner --strip 1 -kC /usr/ && \
+    rm clang+llvm-8.0.0-x86_64-linux-sles11.3.tar.xz
+
+# To build the EBPF code we need headers for Linux 4.9, get them from .deb files from a newer Debian version
+# --skip-old-files (--keep-old-files on older `tar`) makes sure we don't overwrite existing files
+RUN curl -O http://security-cdn.debian.org/debian-security/pool/updates/main/l/linux/linux-headers-4.9.0-9-amd64_4.9.168-1+deb9u3_amd64.deb && \
+    ar p linux-headers-4.9.0-9-amd64_4.9.168-1+deb9u3_amd64.deb data.tar.xz | tar xJ --no-same-owner --keep-old-files -C / && \
+    curl -O http://security-cdn.debian.org/debian-security/pool/updates/main/l/linux/linux-headers-4.9.0-9-common_4.9.168-1+deb9u3_all.deb && \
```
kiyose: This appears to be out of date.  deb9u3 has been replaced with deb9u4 in all urls referenced here.
The new url is:
http://security-cdn.debian.org/debian-security/pool/updates/main/l/linux/linux-headers-4.9.0-9-common_4.9.168-1+deb9u4_all.deb

albertvaka: You are right, and it seems to get updated quite often, so maybe we should host these ourselves somewhere else. This PR needs some re-thinking, though, as using these packages won't even work on CentOS because the files are installed on a different location there. 

I will spend some time trying to fix this tomorrow, since it is blocking you and maybe other people.

PR #9, opened 2019-07-16
Add clang to build images (used by rtload.format)

albertvaka:
Plus reordered some parts of Dockerfiles to keep their layout similar.


rpm-x64/Dockerfile
```diff
@@ -103,9 +104,17 @@ RUN set -ex \
     && ln -s /opt/cmake/bin/cmake /usr/local/bin/cmake \
     && rm cmake.sh
 
+# Install clang and llvm version 8
+# Version built for sles11 is linked against an old enough glibc to be used from this image Versions built for other distros target newer glibcs.
```
truthbk: small typo, missing a `.` before `Versions`.

albertvaka: I've just reworded that

PR #10, opened 2019-07-29
[rpm-x64] make newer gcc (4.7) available

truthbk:
It would be nice to build both the deb and rpm with the same compiler versions.


truthbk:
Note: verified working with a custom build

PR #24, opened 2019-12-20
Adding gettext and autoconf to the builder in order to build procps-ng in omnibus

hush-hush:
New dependencies are needed to build procps-ng latest version. Also autoconf 2.69 is not available for RPM so we have to build it manually.

### DIFF
--- a/rpm-x64/Dockerfile
+++ b/rpm-x64/Dockerfile
@@ -25,7 +25,7 @@ RUN yum -y install \
   which perl-ExtUtils-MakeMaker \
   centos-release-scl pkgconfig \
   curl-devel expat-devel gettext-devel openssl-devel zlib-devel bzip2 \
-  glibc-static tar
+  glibc-static tar libtool
 
 # Git
 RUN curl -OL https://www.kernel.org/pub/software/scm/git/git-2.10.1.tar.gz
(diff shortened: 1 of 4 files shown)


rpm-x64/Dockerfile
```diff
@@ -65,6 +65,17 @@ ENV PKG_CONFIG_LIBDIR $PKG_CONFIG_LIBDIR:$CONDA_PATH/lib/pkgconfig
 RUN conda create -n ddpy2 python python=2
 RUN conda create -n ddpy3 python python=3.7
 
+# Autoconf
+# We need a newer version of autoconf to compile procps-ng (installing 2.69 over 2.63).
+RUN curl -sL -o /tmp/autoconf-2.69.tar.gz https://ftp.gnu.org/gnu/autoconf/autoconf-2.69.tar.gz \
```
truthbk: This is probably fine but will likely overwrite files from the existing `2.63` package. Not sure if this is ideal.... wondering if dropping it in `/usr/local` might be a better idea, and the explicitly calling these autoconf tools from omnibus 🤔

olivielpeau: agreed, @hush-hush could you follow up on this in a future PR?

I'll go ahead and merge the PR as-is now, in time for the freeze

hush-hush: What would be the issue with the overwrite ? We do want to use the latest version by default.

PR #25, opened 2020-01-07
Add BCC in the build images

L3n41c:
This will be needed for the eBPF code of `system-probe`

### DIFF
--- a/system-probe_x64/Dockerfile
+++ b/system-probe_x64/Dockerfile
@@ -1,40 +1,22 @@
 FROM debian:bullseye
 
-ARG BCC_VERSION=v0.12.0
-
 RUN apt update && apt full-upgrade -y
 RUN apt install -y \
         awscli \
+        bison \
         clang \
+        cmake \
+        flex \
         git \
         go-dep \
         golang \
+        libclang-dev \
         libelf-dev \
         linux-headers-amd64 \
         llvm \
         python3-distutils \
         python3-invoke \
         python-pip
-
-RUN apt install -y \
-        bison \
-        cmake \
-        flex \
-        libclang-dev
-
-#     && git clone -b "$BCC_VERSION" --depth=1 https://github.com/iovisor/bcc.git /tmp/bcc \
-#     && mkdir /tmp/bcc/build \
-#     && cd /tmp/bcc/build \
-#     && cmake .. -DCMAKE_INSTALL_PREFIX=/opt/libbcc \
-#     && make -j "$(nproc)" \
-#     && make install \
-#     && cd / \
-#     && rm -rf /tmp/bcc \
-#     && apt remove -y \
-#         bison \
-#         cmake \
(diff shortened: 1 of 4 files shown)


albertvaka:
I would rename `sys-probe` to `system-probe` for consistency.

L3n41c:
@albertvaka : I’ve done the renaming and I’ve asked for the creation of the ECR registries accordingly: DataDog/cloudops#16012.

system-probe_x64/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+        go-dep \
+        golang \
+        libelf-dev \
+        linux-headers-amd64 \
+        llvm \
+        python3-distutils \
+        python3-invoke \
+        python-pip
+
+RUN apt install -y \
```
albertvaka: We will have to disribute libbcc, right? We might want to build it with omnibus (using this image) instead of here, so it gets installed in the prefix where it's going to be used from. Or is there another approach you are going for?

system-probe_x64/Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM debian:bullseye
```
albertvaka: We might want to base this on an older distro to not require a too new glibc.

system-probe_x64/Dockerfile
L3n41c: `libbcc` build has been moved to DataDog/datadog-agent.

system-probe_x64/Dockerfile
L3n41c: The thing is that I already wasted a lot of time trying to build `iovisor/bcc` on an old distro and it was a big fail.
The good news is that there’s already safety belts that are guaranteeing that this is not an issue.
Indeed, what happens in DataDog/datadog-agent#4619 is that:
* `iovisor/bcc` is built inside this `system-probe_$ARCH` docker image environment;
* `system-probe` is built inside this `system-probe_$ARCH` docker image environment (because `iovisor/gobpf` also requires up-to-date glibc and kernel headers);
* all the agent components, included the previously built `libbcc` and `system-probe`, are bundled together with omnibus in the legacy `{rpm,deb}-$ARCH` environment.
And, in this last step, omnibus is checking that everything that is shipped inside `/opt/datadog-agent/embedded` can resolve all its symbols in libs that are also shipped inside `/opt/datadog-agent/embedded`.

If, because of the build process inside `bullseye`, `libbcc` was depending on a glibc symbol not available on `wheezy` or `centos:6`, the build would fail. (Unfortunately, I cannot find the GitLab job where I saw that anymore, but I saw this check)

albertvaka: 👍Thanks! I wasn't sure if omnibus checked each symbol or just the soname. Thanks for taking this into account and for the detailed explanation 🙇

PR #28, opened 2020-02-06
Activate py3 env by default

albertvaka:
So we can call invoke/pip straight away.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -1,6 +1,7 @@
 #!/bin/bash -l
 set -e
 
+source /root/.bashrc && conda activate ddpy3
 eval "$(gimme)"
 
 exec "$@"


albertvaka:
I think the unit tests use python from the image itself. Eg: https://github.com/DataDog/datadog-agent/blob/master/.gitlab-ci.yml#L223

olivielpeau:
oh, you're right 👍 Nevermind then.

FWIW, we used to have separate test vs. build images in the past, but it was a hassle to maintain both.

KSerrania:
This breaks the ARM image at runtime as it does not ship `conda`, we should either have separate `entrypoint.sh`, change the ARM image to use anaconda, or change the x86_64 image to not use anaconda.

albertvaka:
Fix: https://github.com/DataDog/datadog-agent-buildimages/pull/29

PR #34, opened 2020-03-06
Create a symlink for clang in the system probe builder

leeavital:
We recently started pinning clang-8. 

This PR creates a symlink for `clang`  which is the command the `system-probe.build` task expects.

See https://github.com/DataDog/datadog-agent/pull/5079


albertvaka:
D'oh! Thanks :)

albertvaka:
Can you also do this for ARM? https://github.com/DataDog/datadog-agent-buildimages/blob/master/system-probe_arm64/Dockerfile

leeavital:
Whoops, I forgot about all about arm. Done!

PR #37, opened 2020-03-25
Use a pre-compiled version of clang for x64 system-probe builders

leeavital:
We need to keep the versions of clang for the development environment
and build environment in sync because we verify that the checked in
binaries match the binaries from CI.

The easiest way to do this is to use a pre-built version of clang
instead of whatever version is in the apt repositories.

### DIFF
--- a/system-probe_x64/Dockerfile
+++ b/system-probe_x64/Dockerfile
@@ -1,7 +1,6 @@
 FROM debian:bullseye
 
-RUN apt update && apt full-upgrade -y
-RUN apt install -y \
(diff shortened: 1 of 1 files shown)


albertvaka:
Is this check something you will add in another PR? Can you link it here?

leeavital:
@albertvaka we already to the check as part of the agent CI.  You can see it here:

https://github.com/DataDog/datadog-agent/blob/master/.gitlab-ci.yml#L331

albertvaka:
Ok! I'm not a big fan of having a slightly different setup for x64 and arm64 but 🤷‍♀

PR #38, opened 2020-03-26
Add Windows builders

KSerrania:
### What does this PR do?

Adds Windows public Agent builders, built with a newer toolchain than the one we have in old builders.

### Motivation

Use a newer toolchain & open-source the Windows build.

### Open questions

- The old builders embed the embedded python environments that are used for unit tests (presumably to make the tests go faster). Should we put this step here or should it be moved to the unit tests step in the datadog-agent repository?
- I'm not sure that the release step actually works yet (not tested yet), and I'm also not sure that builds with the new builders currently work. Should I make the release jobs manual until we're sure of these two points?


windows/install_embedded_pythons.ps1
```diff
@@ -0,0 +1,79 @@
...
+$py2 = "https://s3.amazonaws.com/dd-agent-omnibus/python-windows-${Env:EMBEDDED_PYTHON_2_VERSION}-amd64.zip"
+$py3 = "https://s3.amazonaws.com/dd-agent-omnibus/python-windows-${Env:EMBEDDED_PYTHON_3_VERSION}-amd64.zip"
+
+if ( $Env:TARGET_ARCH -eq "x86") {
+    $py2 = "https://s3.amazonaws.com/dd-agent-omnibus/python-windows-${Env:EMBEDDED_PYTHON_2_VERSION}-x86.zip"
+    $py3 = "https://s3.amazonaws.com/dd-agent-omnibus/python-windows-${Env:EMBEDDED_PYTHON_3_VERSION}-x86.zip"
+} 
+
+
+$py2Target = "c:\embeddedpy\py2.${Env:EMBEDDED_PYTHON_2_VERSION}"
```
albertvaka: Won't this become `c:\embeddedpy\py2.2.7.17` (note double `2.`)?

KSerrania: That's indeed the case; the directory name doesn't matter though as long as it's different for the two embedded pythons.

derekwbrown:
re: 

> The old builders embed the embedded python environments that are used for unit tests (presumably to make the tests go faster). Should we put this step here or should it be moved to the unit tests step in the datadog-agent repository?

I think we should.  It's a very small increase (%age wise) in the builder size, and makes the tests much faster and more reliable

windows/Dockerfile
```diff
@@ -0,0 +1,94 @@
...
+
+# Install git
+RUN $env:chocolateyUseWindowsCompression = 'true'; cinst -y --no-progress git
+
+# Install 7zip
+RUN $env:chocolateyUseWindowsCompression = 'true'; cinst -y --no-progress 7zip
+
+# Install VS2017
+RUN cinst -y --no-progress visualstudio2017buildtools --params \"--add Microsoft.VisualStudio.ComponentGroup.NativeDesktop.Win81\"
+RUN cinst -y --no-progress visualstudio2017-workload-vctools
```
derekwbrown: we need a very specific set of VS workloads.  how is this workload defined in chocolatey?
Here's what we currently get
https://github.com/DataDog/datadog-agent-builders/blob/db/driverbuild/windows-scripts/install_vstudio.ps1#L16
Also, now need the matching WDK.  Might as well get it now.

windows/Dockerfile
```diff
@@ -0,0 +1,94 @@
...
+
+ARG TARGET_ARCH
+ENV TARGET_ARCH ${TARGET_ARCH:-x64}
+
+ENV GO_VERSION "1.13.8"
+ENV RUBY_VERSION "2.4.3.1"
+ENV PYTHON_VERSION "2.7.17"
+ENV WIX_VERSION "3.11.2"
+
+ENV EMBEDDED_PYTHON_2_VERSION "2.7.17"
```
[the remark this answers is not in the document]
KSerrania: 2.7.17 and 3.8.1 are the versions we currently ship in the Agent, so we should be good there

windows/Dockerfile
albertvaka: We shouldn't block this PR on the Windows Driver Kit, it can be added later.

.gitlab-ci.yml
```diff
@@ -87,3 +110,30 @@ release_rpm_x64:
   extends: .release
   variables:
     IMAGE: rpm_x64
+
+.winrelease: &winrelease
```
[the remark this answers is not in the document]
albertvaka: I think that doesn't work for some reason :(

.gitlab-ci.yml
```diff
@@ -87,3 +115,13 @@ release_rpm_x64:
...
+
+release_windows_x64:
+  extends: .release
+  variables:
+    IMAGE: windows_x64
+
+release_windows_x86:
+  extends: .release
+  variables:
+    IMAGE: windows_x86
```
KSerrania, line 127: Someone else can confirm, but I don't think that will work: we're trying to pull a Windows image on a Linux host, which will probably lead to the following error:
```
image operating system "windows" cannot be used on this platform
```

We probably need to reintroduce a Windows-specific step to push the Windows containers.

windows/install_net35_1909.bat
```diff
@@ -0,0 +1,21 @@
+::
+:: Installs .NET Runtime 3.5, used by Wix 3.11 and the Visual C++ Compiler for Python 2.7
+:: Taken from the mcr.microsoft.com/dotnet/framework/runtime:3.5 Dockerfile:
+:: https://github.com/microsoft/dotnet-framework-docker/blob/abc2ca65b28058f7c71ec8cd8763a8fbf2a9c03f/3.5/runtime/windowsservercore-1909/Dockerfile
+::
+
+set DOTNET_RUNNING_IN_CONTAINER true
+
+curl -fSLo microsoft-windows-netfx3.zip https://dotnetbinaries.blob.core.windows.net/dockerassets/microsoft-windows-netfx3-1909.zip
```
albertvaka, line 9: Why the hell do these depend on the kernel version? 🤦‍♀️

.gitlab-ci.yml
```diff
@@ -63,28 +63,57 @@ build_system-probe_arm64:
...
 
-build_windows_x86:
+build_windows_1909_x86:
   extends: .winbuild
+  tags: [ "runner:windows-docker", "windowsversion:1909" ]
   variables:
     DOCKERFILE: windows/Dockerfile
-    IMAGE: windows_x86
-    BUILDARGS: --build-arg
-    BUILDARGSVAL: TARGET_ARCH=x86
```
albertvaka: 🔥

PR #43, opened 2020-05-14
[macos] Add builder & build scripts for the MacOS Agent

KSerrania:
### What does this PR do?

Adds builder & build scripts for the MacOS Agent, both for unsigned & signed builds.

### DIFF
--- a/macos/README.md
+++ b/macos/README.md
@@ -11,7 +11,7 @@ The scripts in this folder are made to create a build environment for the MacOS 
 ## Contents
 
 - `builder_setup.sh`: installs all required build dependencies
-- `certificate_setup.sh`: installs the developer certificates in keychain & allow automatic access to code-signing applications
+- `certificate_setup.sh`: installs the developer certificates in keychain & allow automatic access to code-signing applications. Only needed if you want to sign the resulting package.
 - `build_script.sh`: does the omnibus build of the Agent
 
 ## Notes on notarization
--- a/macos/README.md
+++ b/macos/README.md
@@ -7,7 +7,7 @@ The scripts in this folder are made to create a build environment for the MacOS 
 
 - A clean MacOS 10.13.6 (High Sierra) host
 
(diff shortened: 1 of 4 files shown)


macos/README.md
```diff
@@ -0,0 +1,43 @@
+# MacOS builder setup scripts
+
+The scripts in this folder are made to create a build environment for the MacOS Agent & create unsigned or signed builds of the MacOS Agent.
+
+
+## Prerequisites
+
+- A clean MacOS 10.13.6 (High Sierra) host
```
albertvaka, line 8: It should theoretically work on newer MacOS. Maybe we can mention that (but say it's not tested)?

macos/README.md
```diff
@@ -0,0 +1,43 @@
...
+- Varying environment variables & files present on the host (see individual files for specific requirements).
+
+## Contents
+
+- `builder_setup.sh`: installs all required build dependencies
+- `certificate_setup.sh`: installs the developer certificates in keychain & allow automatic access to code-signing applications
```
albertvaka, line 14:
```suggestion
- `certificate_setup.sh`: installs the developer certificates in keychain & allow automatic access to code-signing applications. Only needed if you want to sign the resulting package.
```

macos/certificate_setup.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+# Installs the signing certificates and makes sure then can be used
```
albertvaka, line 3:
```suggestion
# Installs the signing certificates and makes sure they can be used
```

macos/README.md
KSerrania: With the current state of the Mac systray app, it won't: we currently target Swift 3 https://github.com/DataDog/datadog-agent/blob/73e744b71aa2b1395e3755ebcc4b0f0ca4105b29/omnibus/config/software/datadog-agent.rb#L226
Recent MacOS 10.14 versions (>= 10.14.4) use XCode 10.2 by default, which removed the ability to target swift 3: https://developer.apple.com/documentation/xcode_release_notes/xcode_10_2_release_notes/swift_5_release_notes_for_xcode_10_2?preferredLanguage=occ
> Swift 3 mode has been removed. Supported values for the -swift-version flag are 4, 4.2, and 5. (43101816).

To be able to run the build with MacOS 10.14 currently, you'd need:
- to find a way to run 10.14.x < 10.14.4, or
- to find a way to install XCode 10.1 instead of the default 10.2 (remark: the brew install will install the Command Line Tools for XCode 10.12, so you'd also need to remove that)

albertvaka: 🙇 Then I think we should clarify that in the comment (ie: the reason we require an old MacOS version).

KSerrania: Added a comment about that

PR #44, opened 2020-05-19
Add ARM32

julien-lebot:


rpm-arm/Dockerfile
```diff
@@ -14,6 +18,7 @@ ENV DD_PIP_VERSION $DD_PIP_VERSION
 ENV DD_SETUPTOOLS_VERSION $DD_SETUPTOOLS_VERSION
 ENV CMAKE_VERSION $CMAKE_VERSION
 ENV CLANG_VERSION $CLANG_VERSION
+ENV GIMME_ARCH $TARGET_ARCH
```
KSerrania, line 21: Isn't it supposed to be `$ARCH`? I don't see any `$GIMME_ARCH` used here, but I do see some `$ARCH` occurrences.

julien-lebot: Good catch, I renamed `$ARCH` to `$TARGET_ARCH` to follow what we do for Windows and to be more clear as to what arch we are talking about (as opposed to host arch, which is 64bit in this case). I will fix the remaining `$ARCH`

PR #45, opened 2020-05-29
[windows] Pin all chocolatey packages

KSerrania:
### What does this PR do?

Pins all chocolatey package versions used in the Windows build to prevent builds from breaking because of a chocolatey package update.

### DIFF
--- a/windows/Dockerfile
+++ b/windows/Dockerfile
@@ -46,6 +46,7 @@ LABEL windows_version=${WINDOWS_VERSION}
 LABEL git_version=${GIT_VERSION}
 LABEL sevenzip_version=${SEVENZIP_VERSION}
 LABEL vs2017buildtools_version=${VS2017BUILDTOOLS_VERSION}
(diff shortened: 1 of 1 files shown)


windows/Dockerfile
```diff
@@ -23,10 +23,17 @@ ENV WINDOWS_VERSION=${WINDOWS_VERSION:-1809}
 ARG TARGET_ARCH
 ENV TARGET_ARCH=${TARGET_ARCH:-x64}
 
+# Chocolatey package versions
+ENV GIT_VERSION "2.26.2"
+ENV SEVENZIP_VERSION "19.0"
+ENV VS2017BUILDTOOLS_VERSION "15.9.23.0"
+ENV VCPYTHON27_VERSION "9.0.0.30729"
```
mx-psi, line 30: Question: why is this not added as a label but the rest are?

KSerrania: Because I forgot it 😄

PR #21, opened 2019-11-07
ARM builders

albertvaka:

PR #49, opened 2020-06-12
Fixes for ARM 32/64

julien-lebot:
This PR introduces a couple of fixes for ARM32/64:
- Conditionally set `GIMME_ARCH` so that go builds don't fail on 64 bit ARM builds
- Update `entrypoint.sh` to use `GIMME_ARCH` for ARM builds so that `eval "$(gimme)"` finds the correct go binary folder
- Use `fakearmv7l` LD_PRELOAD library to patch `uname` at runtime to return architecture `armv7l` for RPM 32 bit build


rpm-arm/Dockerfile
```diff
@@ -18,21 +16,31 @@ ENV DD_PIP_VERSION $DD_PIP_VERSION
...
+RUN rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
+    rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-SIG-AltArch-Arm32
 
 RUN yum groupinstall -y development && yum -y install which perl-ExtUtils-MakeMaker ncurses-compat-libs git procps \
     curl-devel expat-devel gettext-devel openssl-devel systemd-devel zlib-devel bzip2 glibc-static python-devel tar pkgconfig  \
     libtool autoconf policycoreutils-python
 
+# This trick preloads a dynamic library that will replace the uname syscall to pretend the machine is armv7l
+# The fakearmv7l library gets installed in "/usr/local/lib/libfakearmv7l.so"
+RUN git clone https://github.com/DataDog/fakearmv7l ; \
```
mx-psi, line 36: Doesn't seem likely that we will make changes to fakearmv7l but just in case we can do
```suggestion
RUN git clone --depth 1 https://github.com/DataDog/fakearmv7l ; \
```
so that if we ever make changes we don't clone the whole repo history

.gitlab-ci.yml
```diff
@@ -61,22 +61,22 @@ build_deb_armhf:
...
+    DD_TARGET_ARCH: arm32v7
 
 build_rpm_arm64:
   extends: .armbuild
   variables:
-    DOCKERFILE: rpm-arm/Dockerfile
+    DOCKERFILE: rpm-arm64/Dockerfile
     IMAGE: rpm_arm64
-    BASE_IMAGE: arm64v8/centos:7
+    BASE_IMAGE: amazonlinux:2.0.20181114
```
mx-psi, line 71: Why did we change the base image? (just for future reference)

rpm-arm64/Dockerfile
```diff
@@ -0,0 +1,67 @@
...
+RUN gpg --import /gpg-keys/*
+RUN rm -rf /gpg-keys
+RUN curl -sSL https://get.rvm.io | bash -s stable
+RUN /bin/bash -l -c "rvm requirements"
+RUN /bin/bash -l -c "rvm install 2.3 && rvm cleanup all"
+RUN /bin/bash -l -c "gem install bundler --no-document"
+
+# Pip & Invoke
+RUN curl "https://bootstrap.pypa.io/get-pip.py" | python2.7 - pip==${DD_PIP_VERSION} setuptools==${DD_SETUPTOOLS_VERSION}
+RUN pip install invoke distro==1.4.0 awscli==1.16.240
```
mx-psi, line 36: Could we use Python 3 here? Base image doesn't have it so it needs to be installed. Not a blocker but it would useful if we ever need a Python 3 only package.

.gitlab-ci.yml
julien-lebot: Initially it was `amazonlinux`, see #21. Amazon Linux however, doesn't provide 32 bit versions of their image, so I switched to CentOS 7 (which is what Amazon Linux is supposed to be based on I think). However, CentOS 7 is quite old and ships with GCC 4.8.5 - which causes issues when trying to build `rtloader`.
Hence we had two options:
- Install a newer GCC on CentOS 7
- Revert to the Amazon Linux image for 64 bit builds
We chose the latter because we don't build `rtloader` for 32 bit ARM, so it's simpler to revert to the 'known to work' image.

rpm-arm64/Dockerfile
julien-lebot: Good point, I think we could have python 3 in this image, but I think this should be addressed as a separate PR

PR #58, opened 2020-06-30
Update notarization commands

KSerrania:


mx-psi:
I think we need to change step 4 too

Issue #64, opened 2020-07-21
Handle shell appropriately without errorneous "rm /bin/sh && ln -s /bin/bash /bin/sh"


Voronenko:
it is quite unusual to see workarounds like 

```sh
rm /bin/sh && ln -s /bin/bash /bin/sh
```

https://github.com/DataDog/datadog-agent-buildimages/blame/master/deb-x64/Dockerfile#L2

Most likely, what you are looking for is wrapping conda install into  /bin/bash -c "commands"

```Dockerfile

# Update pip, setuptools and misc deps
RUN /bin/bash -c "source /root/.bashrc && conda activate ddpy2 \
    && pip install -i https://pypi.python.org/simple pip==${DD_PIP_VERSION} \
    && pip install --ignore-installed setuptools==${DD_SETUPTOOLS_VERSION} \
    && pip install invoke distro==1.4.0 awscli==1.16.240"

# Update pip, setuptools and misc deps
RUN /bin/bash -c  "source /root/.bashrc && conda activate ddpy3 \
    && pip install -i https://pypi.python.org/simple pip==${DD_PIP_VERSION} \
    && pip install --ignore-installed setuptools==${DD_SETUPTOOLS_VERSION} \
    && pip install invoke distro==1.4.0 awscli==1.16.240"
```

Voronenko:
https://github.com/DataDog/datadog-agent-buildimages/pull/65

mx-psi:
Hi @Voronenko, thanks for your contribution! The PR looks good to me, I think we could get rid of the `bash` calls altogether and use the [`SHELL` command](https://docs.docker.com/engine/reference/builder/#shell) to set the default shell. What do you think? We can handle that ourselves on a future PR otherwise.

PR #65, opened 2020-07-21
improve(deb image) handle unix shell properly

Voronenko:
eliminates erroneous replacing of sh with bash

```Dockerfile
RUN rm /bin/sh && ln -s /bin/bash /bin/sh
```

handles invoking bash properly later in Dockerfile

Fixes  https://github.com/DataDog/datadog-agent-buildimages/issues/64

### DIFF
--- a/deb-x64/Dockerfile
+++ b/deb-x64/Dockerfile
@@ -1,5 +1,4 @@
 FROM debian:wheezy-backports
-RUN rm /bin/sh && ln -s /bin/bash /bin/sh
 
 # Build Args
 ARG GIMME_GO_VERSION=1.13.11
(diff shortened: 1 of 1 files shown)


mx-psi:
Hi @Voronenko, thanks for your contribution! The PR looks good to me, I think we could get rid of the `bash` calls altogether and use the [`SHELL` command](https://docs.docker.com/engine/reference/builder/#shell) to set the default shell. What do you think? We can handle that ourselves on a future PR and I will merge as-is otherwise.

Voronenko:
@mx-psi  I think main issue is replacing sh with bash.  Although usually bash nicely handles syntax that sh can handle, still someone could expect sh to be real sh, and not other shell binary under this name.

I was able to successfully build Dockerfile with current change.

mx-psi:
Great, we will leave as it is! I don't think there is any place that I know of where we use `sh` specific syntax but it's better to be safe than sorry. Thanks again for contributing!

PR #70, opened 2020-08-26
[macos] Add automated notarization script

KSerrania:
### What does this PR do?

Adds automated notarization script. It should work as long as Apple doesn't break the XML output format of `xcrun altool`.

### Motivation

Automate all MacOS build steps.
Prepare automated MacOS builds.


mx-psi (approved):
This LGTM if it has been tested. It's sad that Apple doesn't have an easier way of getting the status code, let's hope this doesn't break often

PR #71, opened 2020-08-27
[macos] Allow customizing the keychain name

KSerrania:
### What does this PR do?

Adds ability to specify on which keychain the signing certificates should be stored & used.
Adds quotes around variables to prevent word splitting.


mx-psi (approved):
LGTM :ship:! I think we could slightly improve this script if we fail if variables are unset but let's leave that for a different PR

PR #78, opened 2020-10-26
[windows] add install script for gcloud sdk

celenechang:
Add a script to install gcloud sdk to the Windows image and add it to the path.

### DIFF
--- a/windows/install_gcloud_sdk.ps1
+++ b/windows/install_gcloud_sdk.ps1
@@ -3,7 +3,8 @@ $ProgressPreference = 'SilentlyContinue'
 [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
 
 Write-Host -ForegroundColor Green Installing Google Cloud SDK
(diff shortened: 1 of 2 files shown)


windows/install_gcloud_sdk.ps1
```diff
@@ -0,0 +1,24 @@
+$ErrorActionPreference = 'Stop'
+$ProgressPreference = 'SilentlyContinue'
+[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
+
+Write-Host -ForegroundColor Green Installing Google Cloud SDK
+$gsdk = 'https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-315.0.0-windows-x86_64.zip'
```
mx-psi, line 6: nit: I would put the SDK version into its own variable, if you think it will change frequently maybe even pass it as an argument so you can modify it easily at build time.

mx-psi (approved):
LGTM, just a nit

mx-psi (approved):
LGTM! You should merge/rebase master to double check that CI passes

mx-psi:
Turns out CI is not going to pass even if you merge 😞 we have to wait until unrelated fixes are made until we can build an image with these changes so I think it's best to wait for merging this until that happens

PR #80, opened 2020-10-28
[windows] Update msys distribution

KSerrania:
### What does this PR do?

Updates the msys distribution to the latest available: packages for previous distributions have been pruned from the msys repositories (see https://www.msys2.org/news/).

Uses `ridk install 2` to force-update to the latest packages. The packages provided in the latest msys release are broken (cmake fails to use the `cc` compiler), so we have to get the fixed packages by upgrading to the latest available packages.


windows/Dockerfile
```diff
@@ -132,7 +132,10 @@ RUN gem install bundler
 
 # Install msys2 system & install 64-bit C/C++ compilation toolchain
 RUN cinst -y --no-progress msys2 --params \"/NoUpdate\" --version $ENV:MSYS_VERSION
-RUN ridk install 3
+
+# TODO: as soon as a new msys distribution (with working packages) is available, use it
+# and stop doing ridk install 2, which force-upgrades packages to the latest available.
+RUN ridk install 2 3
```
julien-lebot, line 138: Does `ridk install 2` upgrade MSYS to the latest version ? I.e. is it the same as running `cinst msys2` without the `NoUpdate` parameter ?

julien-lebot: If we're going to remove "2" in the future, then the current PR is fine by me as it is.

julien-lebot (approved):

PR #86, opened 2020-12-03
[deb-x64, rpm-x64, macos] Update IBM MQ library

KSerrania:
### What does this PR do?

Updates IBM MQ library to 9.2.0.0 on MacOS. Also updates it to 9.2.0.1 on Linux to match.

### Motivation

The current version (9.1.5.0) is not available anymore for MacOS.


mx-psi (approved):
LGTM, but I think it's better if we merge #87 first so that we can test rpm

PR #93, opened 2020-12-14
Fix winrelease target; needs to run on newest available version

derekwbrown:
Fixes oversight in previous commit.


derekwbrown:
Reopened new PR; started branch from wrong place.

PR #100, opened 2020-12-21
Add docker login to build job

albertvaka:
Since it also pulls images from Dockerhub during `docker build`.


KSerrania (commented):
LGTM, I think we want to have the same thing in the `.armbuild` job for ARM images?

KSerrania (approved):

PR #101, opened 2021-01-04
Add 20H2 builders, update msys2 to 20201109

derekwbrown:
Adds 20H2 builders to the set of windows builders.
This commit also changes the MSYS version, which fixes the problem that the MSYS install intermittently hangs at build time.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -239,6 +239,16 @@ build_windows_2004_x64:
     DD_TARGET_ARCH: x64
     WINDOWS_VERSION: 2004
 
+build_windows_20h2_x64:
+  extends: .winbuild
+  tags: [ "runner:windows-docker", "windowsversion:20h2" ]
+  variables:
+    DOCKERFILE: windows/Dockerfile
+    IMAGE: windows_20h2_x64
+    BASE_IMAGE: mcr.microsoft.com/dotnet/framework/runtime:4.8-windowsservercore-20H2
+    DD_TARGET_ARCH: x64
+    WINDOWS_VERSION: 20H2
+
 .test_windows:
   extends: .test
   timeout: 2h 00m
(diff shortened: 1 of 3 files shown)


windows/install_net35_20h2.bat
```diff
@@ -0,0 +1,21 @@
+::
+:: Installs .NET Runtime 3.5, used by Wix 3.11 and the Visual C++ Compiler for Python 2.7
+:: Taken from the mcr.microsoft.com/dotnet/framework/runtime:3.5 Dockerfile:
+;; https://github.com/microsoft/dotnet-framework-docker/blob/e0b59f4aeb47bd6bf13e4c7ec6676a1935306df9/src/runtime/3.5/windowsservercore-20H2/Dockerfile
```
KSerrania, line 4: This comment should start with `::`. This line currently causes an error:
```
C:\>https://github.com/microsoft/dotnet-framework-docker/blob/e0b59f4aeb47bd6bf13e4c7ec6676a1935306df9/src/runtime/3.5/windowsservercore-20H2/Dockerfile
'https:' is not recognized as an internal or external command,
operable program or batch file.
```

(which doesn't make the script fail, since we don't exit on error here)

KSerrania (commented):
Could the PR name be updated to show that this updates the msys distro, so that we can see this in git logs?

.gitlab-ci.yml
```diff
@@ -394,3 +412,9 @@ release_windows_2004_x64:
   needs: [ "test_windows_2004_x64" ]
   variables:
     IMAGE: windows_2004_x64
+
+release_windows_20h2_x64:
+  extends: .winrelease
+  needs: [ "test_windows_20h2_x64" ]
+  variables:
+    IMAGE: windows_202h_x64
```
KSerrania, line 420:
```suggestion
    IMAGE: windows_20h2_x64
```

KSerrania (approved):
LGTM once the small typo in the release job is fixed

PR #105, opened 2021-01-08
log builder instance id for easier debugging

derekwbrown:


KSerrania (commented):
LGTM, can you also add it to the test jobs (`.test_windows`)?

KSerrania (approved):

PR #106, opened 2021-01-11
Use a single DockerHub repo for all Windows images

albertvaka:
Use a tag per kernel version (and no `latest` tag, since it no longer makes sense), so we don't have to create N repos.

Also:
 - Removes the public x86 images push, since we don't support those.
 - Adds a missing `rmi` for the private repos' `latest` tag, which could cause old images to pile up in the builders.


derekwbrown:
do I understand this correctly, that now to use images from docker hub, I can't just get latest any more.
Furthermore, I now have to know the exact git hash (essentially), and do something like
`docker pull datadog/agent-buildimages-windows_x64:2004-(somecommithash)`
in order to get the 2004 builder, instead of what I would do today which is simply
`docker pull datadog/agent-buildimages-2004`
and be done?

albertvaka:
Good point. I've changed it so we also tag the latest for each kernel version as `agent-buildimages-windows_x64:<kernel_version>`, so you can now pull `agent-buildimages-windows_x64:2004` and that will be the latest `2004` image we released. That is similar to what they do at https://hub.docker.com/_/microsoft-powershell and should fix the convenience problems.

derekwbrown (approved):

PR #110, opened 2021-02-05
Fix system-probe build

albertvaka:
- Bump llvm to 11.0.1 (11.0.0 is not available for Ubuntu 16.04)
- Download clang to build system-probe in the tests

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -12,6 +12,9 @@ variables:
   # Thus, to release buildimages, we do the same thing as what we do in the Agent: we run the Docker publish script in
   # the buildimage for the highest Windows version supported.
   WINDOWS_RELEASE_IMAGE: 486234852809.dkr.ecr.us-east-1.amazonaws.com/ci/datadog-agent-buildimages/windows_2004_x64:v3541301-fcf9226
+  S3_CP_CMD: aws s3 cp $S3_CP_OPTIONS
+  S3_PERMANENT_ARTIFACTS_URI: s3://dd-ci-persistent-artefacts-build-stable/datadog-agent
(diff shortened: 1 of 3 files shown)


.gitlab-ci.yml
```diff
@@ -162,6 +165,9 @@ test_rpm_arm64:
     - git clone https://github.com/DataDog/datadog-agent /go/src/github.com/DataDog/datadog-agent && cd /go/src/github.com/DataDog/datadog-agent
     - git checkout master
     - inv -e deps --verbose
+    - $S3_CP_CMD $S3_PERMANENT_ARTIFACTS_URI/clang-$ARCH-11.0.0.tar.xz /tmp/clang.tar.xz
```
mx-psi, line 168: I have close to zero context on what this does or how it works, but I found it weird that we use clang version `11.0.0` on the tests but we use version `11.0.1` on the Dockerfile. Is there a reason for that? Is that fine?

albertvaka: For some reason that I don't know (I'll try to find an answer after writing this), there are two clangs in play when building the system-probe: one we built ourselves and lives in S3, and the official one we download. I agree it's a bit weird that we need both, and even more weird that one comes in the image and the other doesn't. And it's true that upgrading one and not the other might have unintended consequences... so let's figure this out before merging this PR.

albertvaka: Got my answer:
> The one we build ourselves only has the BPF backend and ends up being built statically with the system-probe. It is used for runtime compilation. The one in the image is used for building the ebpf object files ahead of time (rather than at runtime).
Given that I don't think that having different versions is a problem, but I will still update the one built by us, so they match.

mx-psi (approved):

PR #111, opened 2021-02-05
Use conda in ARM builders to get a newer python3

albertvaka:
- Deduplicates the code to install conda and set up python by moving it to its own script.
- Stops installing python from the distro repos in ARM builders, installs with conda instead.
- Makes `DD_TARGET_ARCH` available on all builders and consistent.
- Unifies arm and x64 test jobs template on gitlab, since they both use conda now.
- Bumps the patch version of miniconda.

I had to use a fork of miniconda that supports ARM since the official one doesn't.


setup_python.sh
```diff
@@ -0,0 +1,40 @@
...
+
+# Setup pythons
+conda create -n ddpy2 python python=2
+conda create -n ddpy3 python python=3.8
+
+# Update pip, setuptools and misc deps
+conda activate ddpy2 \
+    && pip install -i https://pypi.python.org/simple pip==${DD_PIP_VERSION} \
+    && pip install --ignore-installed setuptools==${DD_SETUPTOOLS_VERSION} \
+    && pip install invoke distro==1.4.0 awscli==1.16.240
```
mx-psi, line 33: `invoke` wasn't pinned before but I think we should pin it (on `datadog-agent` we use `1.4.1`):
```suggestion
    && pip install invoke==1.4.1 distro==1.4.0 awscli==1.16.240
```

setup_python.sh
```diff
@@ -0,0 +1,40 @@
...
+
+# Update pip, setuptools and misc deps
+conda activate ddpy3 \
+    && pip install -i https://pypi.python.org/simple pip==${DD_PIP_VERSION} \
+    && pip install --ignore-installed setuptools==${DD_SETUPTOOLS_VERSION} \
+    && pip install invoke distro==1.4.0 awscli==1.16.240
```
mx-psi, line 39:
```suggestion
    && pip install invoke==1.4.1 distro==1.4.0 awscli==1.16.240
```

mx-psi (approved):

PR #112, opened 2021-02-15
Add VCPKG

julien-lebot:
This PR installs VCPKG as an alternative to Nuget for C++ dependencies management.


windows/install_vcpkg.ps1
```diff
@@ -0,0 +1,4 @@
+git clone https://github.com/microsoft/vcpkg
```
mx-psi, line 1: We can do a shallow clone here
```suggestion
git clone --depth 1 https://github.com/microsoft/vcpkg
```

windows/install_vcpkg.ps1
```diff
@@ -0,0 +1,4 @@
+git clone --depth 1 https://github.com/microsoft/vcpkg
```
albertvaka, line 1: Can we pin a release to install, instead of cloning from master? For build consistency.

albertvaka (approved):

PR #125, opened 2021-04-08
Hack to fix python2 unit tests in Amazon Linux

albertvaka:


setup_python.sh
```diff
@@ -55,3 +55,7 @@ conda activate ddpy3 \
     && pip install --ignore-installed setuptools==${DD_SETUPTOOLS_VERSION} \
     && pip install invoke==1.4.1 distro==1.4.0 awscli==1.16.240
 
+if [ "$DD_TARGET_ARCH" = "aarch64" ] ; then
+    # Conda creates "lib" but on Amazon Linux, the embedded Python2 we use in unit tests will look in "lib64" instead
+    ln -s "${CONDA_PATH}/envs/ddpy2/lib" "${CONDA_PATH}/envs/ddpy2/lib64"
```
KSerrania, line 60: Out of curiosity: why isn't this needed for the Python 3 env? Is it because `conda` does create `lib64` in that case?

KSerrania (approved):

setup_python.sh
albertvaka: No, it only creates `lib` Python3 seems happy with that 🤷

PR #130, opened 2021-04-21
Use own vcpython27

julien-lebot:
This PR uses the VCPython27 from our S3 bucket.

### DIFF
--- a/windows/install_vcpython.ps1
+++ b/windows/install_vcpython.ps1
@@ -6,12 +6,18 @@ $ErrorActionPreference = 'Stop'
 [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
 
 
-$vcpmsi = 'https://download.microsoft.com/download/7/9/6/796EF2E4-801B-4FC4-AB28-B59FBF6D907B/VCForPython27.msi'
+$vcpmsi = 'https://s3.amazonaws.com/dd-agent-omnibus/VCForPython27.msi'
 
 Write-host -ForegroundColor Green Downloading VC++ for Python
 $out = "$($PSScriptRoot)\vcp.msi"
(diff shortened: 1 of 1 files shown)


derekwbrown (approved):

windows/install_vcpython.ps1
```diff
@@ -6,12 +6,18 @@ $ErrorActionPreference = 'Stop'
...
 
-$vcpmsi = 'https://download.microsoft.com/download/7/9/6/796EF2E4-801B-4FC4-AB28-B59FBF6D907B/VCForPython27.msi'
+$vcpmsi = 'https://s3.amazonaws.com/dd-agent-omnibus/VCForPython27.msi'
 
 Write-host -ForegroundColor Green Downloading VC++ for Python
 $out = "$($PSScriptRoot)\vcp.msi"
 
 (New-Object System.Net.WebClient).DownloadFile($vcpmsi, $out)
+
+if ((Get-FileHash -Algorithm MD5 $out).Hash -ne "4E6342923A8153A94D44FF7307FCDD1F")
```
albertvaka, line 16: Please switch to SHA256 or something else. MD5 is quite broken.

albertvaka (approved):

windows/install_vcpython.ps1
julien-lebot: I used MD5 because that's what was used in the Chocolatey package https://community.chocolatey.org/packages/vcpython27
This way we know it's the same as the Chocolatey one.
MD5 is quite broken, but we're only checking file integrity, so does it really matter ?

albertvaka: It could happen that the file is changed in such a way that it produces the same MD5 (that's relatively easy to do).

PR #132, opened 2021-05-04
Make `inv deps` not verbose on tests

mx-psi:
Follow up to DataDog/datadog-agent#7990


julien-lebot (approved):
LGTM Much easier to read the output !

PR #134, opened 2021-05-14
Bump conda version

albertvaka:
Should fix it getting stuck on "Unpacking payload" sometimes.

Conda issue: https://github.com/ContinuumIO/anaconda-issues/issues/11466


deb-arm/Dockerfile
```diff
@@ -5,7 +5,7 @@ FROM ${BASE_IMAGE}
 
 # Build Args
 ARG GIMME_GO_VERSION=1.15.11
-ARG DD_CONDA_VERSION=4.7.12-0
+ARG DD_CONDA_VERSION=4.8.3-0
```
mx-psi, line 8: Why this specific version instead of the last one (4.10.1)?

albertvaka: Because mistakes were made, let me change it to the latest.

mx-psi: Okay looks like 4.10 says 

> NOTE: This release formally drops support for Python 2.7 and Python < 3.6.

but we can use the last 4.9.x

albertvaka: I checked here and the latest one listed is 4.8: https://docs.conda.io/en/latest/miniconda_hashes.html

Where did you get your version number from?

albertvaka: Ok, there's a 4.10 version for conda-forge: https://github.com/conda-forge/miniforge/releases/
But 4.9 is the latests for miniconda: https://repo.anaconda.com/miniconda/

albertvaka: I've updated the PR to use the latest of each, and moved the version number definition to the `setup_python.sh` script so we can change it from a single place. 

This adds a potential difference between the arm and x64 builders, but on the other hand if conda-forge released 4.10 ahead of miniconda it might be because of a bug in 4.9, so I think this is better.

albertvaka: Oh, I didn't see your comment about 4.10 🤦

setup_python.sh
```diff
@@ -9,9 +9,11 @@ function detect_distro(){
 
 case $DD_TARGET_ARCH in
 "x64")
+    DD_CONDA_VERSION=4.9.2
     CONDA_URL=https://repo.anaconda.com/miniconda/Miniconda3-${DD_CONDA_VERSION}-Linux-x86_64.sh
```
mx-psi, line 13: I think we have to do this?
```suggestion
    CONDA_URL=https://repo.anaconda.com/miniconda/Miniconda3-py39_${DD_CONDA_VERSION}-Linux-x86_64.sh
```

mx-psi (approved):

PR #135, opened 2021-05-24
[windows] Add WebBuildTools Visual Studio Workload to build image

andrewlock:
This installs the web build tools as part of the Visual Studio installation, as required by the .NET APM team. 

It also makes a slight performance optimisation to the .NET Core installation to avoid NuGet document extraction, and to prime the "first-time experience" so that this doesn't need to run on every build

### DIFF
--- a/windows/install_dotnetcore.ps1
+++ b/windows/install_dotnetcore.ps1
@@ -14,7 +14,14 @@ $out = "$($PSScriptRoot)\dotnetcoresdk.exe"
 Write-Host -ForegroundColor Green Downloading $Url to $out
 (New-Object System.Net.WebClient).DownloadFile($Url, $out)
 
+# Skip extraction of XML docs - generally not useful within an image/container - helps performance
+setx NUGET_XMLDOC_MODE skip
+
 start-process -FilePath $out -ArgumentList "/install /quiet /norestart" -wait
 
 Remove-Item $out
+
+# Trigger first run experience by running arbitrary cmd
+dotnet help
+
 Write-Host -ForegroundColor Green Done with DotNet Core SDK $Version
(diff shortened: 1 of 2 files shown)


derekwbrown (approved):

windows/install_dotnetcore.ps1
```diff
@@ -14,7 +14,14 @@ $out = "$($PSScriptRoot)\dotnetcoresdk.exe"
...
 
+# Skip extraction of XML docs - generally not useful within an image/container - helps performance
+setx NUGET_XMLDOC_MODE skip
+
 start-process -FilePath $out -ArgumentList "/install /quiet /norestart" -wait
 
 Remove-Item $out
+
+# Trigger first run experience by running arbitrary cmd
+dotnet help
```
julien-lebot, line 25: Can/should we disable the first run experience altogether ? They seem to suggest to use DOTNET_NOLOGO [here.](https://github.com/dotnet/sdk/issues/12195)

julien-lebot: Also, I know that dotnet also compiles assemblies into native format and caches them in the GAC the first time they run - would it be interesting to do that in the builder ?

julien-lebot (commented):
Just a couple of clarification questions.

windows/install_dotnetcore.ps1
andrewlock: Thanks for the questions @julien-lebot!

> Can/should we disable the first run experience altogether

I thought it was probably best to run the first time experience, as this unpacks the NuGet packages into the cache. That doesn't make sense to do in GitHub actions (as in that link for example) but for us, priming the cache should actually be good for build times 🙂 

> I know that dotnet also compiles assemblies into native format and caches them in the GAC the first time they run

.NET Core doesn't use the GAC, so I don't believe that applies here? I believe the images should already be NGEN'd so I don't think there's anything more for us to do?

julien-lebot (approved):
LGTM, thanks for the PR !

PR #143, opened 2021-06-18
Add rust 1.60

albertvaka:
Rust is required to build some python dependencies. For dependencies with binary wheels that target manylinux2010 we can simply use the binary wheel and not build the dependency ourselves, but for wheels that target manylinux2014 we need to do the build (and thus need Rust).


deb-x64/Dockerfile
```diff
@@ -104,6 +104,10 @@ RUN curl -Sl -O https://dd-agent-omnibus.s3.amazonaws.com/kernel-4.9-headers-deb
 # Download and install golangci-lint
 RUN curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $GOPATH/bin v1.21.0
 
+# Rust is needed to compile some python libs
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain 1.41.0
```
mx-psi, line 108:
```suggestion
RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain ${RUST_VERSION}
```

Also, why not use 1.53 (and bump the version on builders too)?

albertvaka: Because this was an old branch I already had :) I've bumped it to 1.53 like you suggested but this might not be the solution we go for since rust can't be installed in suse11.

mx-psi (approved):

deb-x64/Dockerfile
```diff
@@ -138,6 +140,14 @@ RUN curl -sSfL -o golangci-lint-install.sh https://raw.githubusercontent.com/gol
     && echo "942efc00799b83aebaf6628c34e1151467b7fa21f0cfd78554279dc73f8c8ef3  $GOPATH/bin/golangci-lint" | sha256sum --check \
     && rm golangci-lint-install.sh
 
+# Rust is needed to compile some python libs
+RUN curl -sSL -o rustup-init.sh https://raw.githubusercontent.com/rust-lang/rustup/master/rustup-init.sh \
```
mx-psi, line 144: The script [changes from time to time](https://github.com/rust-lang/rustup/commits/master/rustup-init.sh), should we pin a specific commit/tag instead of `master`?

KSerrania: Yes, we should. As an example, we can reuse what's done in the [SUSE image build](https://github.com/DataDog/datadog-agent-buildimages/blob/dfc7750f76243b754297f4835dd82aa440aa8efb/suse-x64/Dockerfile#L131-L137).

albertvaka: Dang, I changed to using the script from master just so we could pin it, then forgot to pin it.

mx-psi (approved):

deb-x64/Dockerfile
```diff
@@ -138,6 +140,13 @@ RUN curl -sSfL -o golangci-lint-install.sh https://raw.githubusercontent.com/gol
     && echo "942efc00799b83aebaf6628c34e1151467b7fa21f0cfd78554279dc73f8c8ef3  $GOPATH/bin/golangci-lint" | sha256sum --check \
     && rm golangci-lint-install.sh
 
+# Rust is needed to compile some python libs
+RUN curl -sSL -o rustup-init.sh https://raw.githubusercontent.com/rust-lang/rustup/1.24.3/rustup-init.sh \
+    && echo "a3cb081f88a6789d104518b30d4aa410009cd08c3822a1226991d6cf0442a0f8  rustup-init.sh" | sha256sum --check \
+    && sh rustup-init.sh -y --default-toolchain ${RUST_VERSION} \
+    && echo "3dc5ef50861ee18657f9db2eeb7392f9c2a6c95c90ab41e45ab4ca71476b4338  $HOME/.cargo/bin/rustc" | sha256sum --check
```
KSerrania, line 147: One small nit: if the `RUST_VERSION` is configurable, then the sha256 should also be, otherwise this check will fail on any `RUST_VERSION` that's not `1.60.0`

albertvaka: IMO `RUST_VERSION` is as configurable as the sha as is, but I can move the SHA to the top.

albertvaka: I don't think anyone is using the `ARGs` to the Dockerfile anyway and would rather remove them, maybe leaving some things as `ENV` so they can be reused.

KSerrania: I see your point. For this PR let's leave everything in a consistent state, and then we can discuss removing all `ARG`s from the Dockerfile.

PR #146, opened 2021-07-09
[macos] Update cmake & add -e to build scripts

KSerrania:
### What does this PR do?

Adds `-e` to the MacOS scripts to make them fail early.

Updates the `cmake` version we install in the builder to a version that doesn't depend on `sphinx-doc` (see [this formula](https://github.com/DataDog/homebrew-datadog-agent-macos-build/blob/master/Formula/cmake%403.18.2.2.rb)), as this dependency fails to install with `Given no hashes to check 169 links for project 'pip': discarding no candidates`.

Example run before change: https://github.com/DataDog/datadog-agent-macos-build/runs/3018046927?check_suite_focus=true
Example run after change: https://github.com/DataDog/datadog-agent-macos-build/runs/3027897093?check_suite_focus=true

### DIFF
--- a/macos/builder_setup.sh
+++ b/macos/builder_setup.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -e
 
 # Setups a MacOS builder that can do unsigned builds of the MacOS Agent.
 # The .build_setup file is populated with the correct envvar definitions to do the build,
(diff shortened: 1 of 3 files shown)


macos/builder_setup.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -e
```
albertvaka, line 1: nit: This is ignored if you run the scripts by doing `bash ./my_script.sh`, while `set -e` works in both cases. Probably not an issue here, but in general I think the later is safer.

albertvaka (approved):
✅ Left a minor comment but feel free to ignore.

macos/builder_setup.sh
KSerrania: I think it does matter here, in `datadog-agent-macos-build` we do `bash ./script.sh`. However, I tried a run with `set -e`, and it seems we have some non-critical errors that we'd need to fix (see https://github.com/DataDog/datadog-agent-macos-build/runs/3029648487?check_suite_focus=true) that are likely due to a caching issue.

I'll add a `FIXME` here and in `datadog-agent-macos-build` to explain why we can't set `set -e` right now.

PR #148, opened 2021-07-23
Update IBM MQ client to 9.2.2.0

KSerrania:
The previous version got removed, so we now have to update to 9.2.2.0.

Additionally, only `.pkg` packages are available on MacOS now: https://public.dhe.ibm.com/ibmdl/export/pub/software/websphere/messaging/mqdev/mactoolkit/, so the way to install the package on MacOS was changed to use that.


albertvaka (approved):
✅

PR #155, opened 2021-09-07
Remove unnecessary and buggy `&&`-chaining

djmitche:
The presence of `set -e` already implies that if a command fails, the
script will stop.  So there's no need to chain comamnds together with
`&&`.  And, such `&&` chains interact poorly with `set -e` and can cause
errors to be ignored.

https://gitlab.ddbuild.io/DataDog/datadog-agent-buildimages/-/jobs/83605383 shows a _failed_ run that was still green.  It failed to install pip==19.1, and as a consequence didn't install invoke.

```shell
set -ex
false && true
echo UHOH
```
output:
```
+ false
+ echo UHOH
```

### DIFF
--- a/setup_python.sh
+++ b/setup_python.sh
@@ -50,16 +50,16 @@ conda create -n ddpy2 python python=2
 conda create -n ddpy3 python python=$PY3_VERSION
 
 # Update pip, setuptools and misc deps
-conda activate ddpy2 \
-    && pip install -i https://pypi.python.org/simple pip==${DD_PIP_VERSION} \
-    && pip install --ignore-installed setuptools==${DD_SETUPTOOLS_VERSION} \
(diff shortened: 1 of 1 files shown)


julien-lebot (approved):
Thanks you for taking a look at this issue, code LGTM assuming the pipeline passes.

I didn't know about this but the man page is quite explicit:
```
-e      Exit  immediately if a simple command (see SHELL GRAMMAR
        above) exits with a non-zero status.  The shell does not
        exit  if  the  command that fails is part of the command
        list immediately following a  while  or  until  keyword,
        part  of the test in an if statement, part of a && or ││
        list, or if the command’s return value is being inverted
        via  !.   A  trap on ERR, if set, is executed before the
        shell exits.
```

albertvaka (approved):

PR #159, opened 2021-09-09
Simplify to a single conda install command

brycekahle:
This appears to fix the nikos build.

### DIFF
--- a/omnibus-nikos_x64/Dockerfile
+++ b/omnibus-nikos_x64/Dockerfile
@@ -60,8 +60,7 @@ COPY ./conda.sh /etc/profile.d/conda.sh
 ENV PATH "${CONDA_PATH}/bin:${PATH}"
 ENV PKG_CONFIG_LIBDIR "${PKG_CONFIG_LIBDIR}:${CONDA_PATH}/lib/pkgconfig"
 
-RUN conda install -c psi4 gcc-5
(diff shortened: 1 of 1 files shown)


ISauve:
Thanks for the fix @brycekahle! Looks okay to me, but I'll wait for the image to be built & then trigger an `omnibus-nikos` build on it to make sure everything's okay before giving it a ✅

ISauve (dismissed):
This causes the `omnibus-nikos` build to fail with the following error: [link](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/84463924)
```
Error:
    /root/miniconda3/compiler_compat/ld: cannot find -lpthread
/root/miniconda3/compiler_compat/ld: cannot find -lc
collect2: error: ld returned 1 exit status
error: command '/root/miniconda3/bin/gcc' failed with exit code 1
```
I'm looking into a solution now.

ISauve:
So the problem was an incompatibility between Python & OpenSSL 1.1.1l. I've fixed the issue (a bit hackily, since unfortunately the gcc-5 install seems to always upgrade openssl back to the incompatible version, so I just made sure to perform that step last) & removed this build image's dependency on the shared `setup_python` script so that any future changes there do not affect this image.

ISauve:
Closing in favour of #162

PR #160, opened 2021-09-10
[deb_x64] Build go1.16 from source

KSerrania:
### What does this PR do?

Builds `go1.16` from source in the `deb_x64` builder, using `gcc 4.9` and `go1.15.11`.

### Motivation

With the publicly distributed `go1.16` binaries, some weird linker errors appear. For instance, `inv -e install-tools` fails with:
```
/root/.gimme/versions/go1.16.7.linux.amd64/pkg/tool/linux_amd64/link: running gcc failed: exit status 1
/usr/bin/ld: /tmp/go-link-101083533/000010.o(.text+0x74): unresolvable AWAVAUATUSH��h�*H�T$(������ relocation against symbol `stderr@@GLIBC_2.2.5'
/usr/bin/ld: BFD (GNU Binutils for Debian) 2.22 internal error, aborting at ../../bfd/reloc.c line 443 in bfd_get_reloc_size
/usr/bin/ld: Please report this bug.
```

According to https://github.com/golang/go/issues/43996#issuecomment-782547917, this is due to the builder having a lower `binutils` version than what was used to build the publicly distributed go binaries.

### DIFF
--- a/setup_go.sh
+++ b/setup_go.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+set -ex
+
+# Install gimme to get go1.15
+curl -sL -o /bin/gimme https://raw.githubusercontent.com/travis-ci/gimme/master/gimme
+chmod +x /bin/gimme
+eval $(gimme 1.15.11)
+
+# Install gcc 4.9 (needed to compile go, as it introduces -fdiagnostics-color which is required
+# to compile go)
+apt-get update && apt-get install -y gcc-4.9-backport
+
+git clone https://go.googlesource.com/go goroot && cd goroot
+git checkout go$GO_VERSION
+cd src 
+
+# Use the bundled clang + go 1.15.11 to build the target go version
+CC=/usr/lib/gcc-4.9-backport/bin/gcc ./all.bash
+
+# Update PATH to include the built go binaries
+echo 'export PATH="/goroot/bin:$PATH"' >> /root/.bashrc
--- a/setup_go.sh
+++ b/setup_go.sh
@@ -3,7 +3,9 @@
 set -ex
 
 # Install gimme to get go1.15.11
-curl -sL -o /bin/gimme https://raw.githubusercontent.com/travis-ci/gimme/master/gimme
+curl -sL -o /bin/gimme https://raw.githubusercontent.com/travis-ci/gimme/v1.5.4/gimme
+# Check sha256sum, fail otherwise
+echo "03b295636d4e22870b6f6e9bc06a71d65311ae90d3d48cbc7071f82dd5837fbc  /bin/gimme" | sha256sum --check
 chmod +x /bin/gimme
 eval "$(gimme 1.15.11)"
 
@@ -12,8 +14,7 @@ eval "$(gimme 1.15.11)"
 # We don't add it to the PATH so that it's not used in the Agent / rtloader builds
 apt-get update && apt-get install -y gcc-4.9-backport
 
-git clone --branch "go$GO_VERSION" --depth 1 https://go.googlesource.com/go goroot && cd goroot
-cd src 
+git clone --branch "go$GO_VERSION" --depth 1 https://go.googlesource.com/go goroot && cd goroot/src
 
 # Use gcc 4.9 + go1.15.11 to build the target go version
 CC=/usr/lib/gcc-4.9-backport/bin/gcc ./all.bash
--- a/setup_go.sh
+++ b/setup_go.sh
@@ -10,24 +10,24 @@ chmod +x /bin/gimme
 eval "$(gimme 1.15.11)"
 
 # Install gcc 4.9 (needed to compile go, as it introduces -fdiagnostics-color which is required
-# in the post-compilation steps of go)
-# We don't add it to the PATH so that it's not used in the Agent / rtloader builds
+# in the post-compilation steps of go).
+# Note: this doesn't replace the existing gcc (4.7.2), gcc 4.9 is installed in
+# /usr/lib/gcc-4.9-backport/bin.
 apt-get update && apt-get install -y gcc-4.9-backport
 
 git clone --branch "go$GO_VERSION" --depth 1 https://go.googlesource.com/go goroot && cd goroot/src
 
 # Use gcc 4.9 + go1.15.11 to build the target go version
-CC=/usr/lib/gcc-4.9-backport/bin/gcc ./all.bash
+# HACK: cgo tries to look for gcc in the same place that CC
+# pointed to when go was compiled. By default, CC=gcc.
+# We don't want to keep gcc 4.9 after building go, so we temporarily
+# change the PATH so that "gcc" is the gcc 4.9 we installed.
+# Then, when cgo is used, it will use "gcc" which will be gcc 4.7.
+export PATH="/usr/lib/gcc-4.9-backport/bin:$PATH"
+./all.bash
 
 # Remove gcc 4.9 after building go
 apt-get remove -y gcc-4.9-backport
-
-# HACK: cgo tries to look for gcc in the same place that CC
-# pointed to when go was compiled. By default, this is fine (since CC=gcc),
-# but here since CC is more specific, cgo will try to find gcc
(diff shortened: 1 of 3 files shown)


setup_go.sh
```diff
@@ -0,0 +1,23 @@
...
+# in the post-compilation steps of go)
+# We don't add it to the PATH so that it's not used in the Agent / rtloader builds
+apt-get update && apt-get install -y gcc-4.9-backport
+
+git clone https://go.googlesource.com/go goroot && cd goroot
+git checkout go$GO_VERSION
+cd src 
+
+# Use gcc 4.9 + go1.15.11 to build the target go version
+CC=/usr/lib/gcc-4.9-backport/bin/gcc ./all.bash
```
mx-psi, line 20: `make.bash` has these two variables. If we don't set `CC_FOR_TARGET`, does that mean that we use `gcc` 4.9 for cgo?

```
# CC: Command line to run to compile C code for GOHOSTARCH.
# Default is "gcc". Also supported: "clang".
#
# CC_FOR_TARGET: Command line to run to compile C code for GOARCH.
# This is used by cgo.  Default is CC.
```

setup_go.sh
```diff
@@ -0,0 +1,23 @@
...
+curl -sL -o /bin/gimme https://raw.githubusercontent.com/travis-ci/gimme/master/gimme
+chmod +x /bin/gimme
+eval $(gimme 1.15.11)
+
+# Install gcc 4.9 (needed to compile go, as it introduces -fdiagnostics-color which is required
+# in the post-compilation steps of go)
+# We don't add it to the PATH so that it's not used in the Agent / rtloader builds
+apt-get update && apt-get install -y gcc-4.9-backport
+
+git clone https://go.googlesource.com/go goroot && cd goroot
```
mx-psi, line 15: Should we delete `goroot` (except for `/goroot/bin`) once the build is finished? (Or, alternatively, do this on a separate build stage and just copy the binary)

KSerrania: I can try doing this. However, `gimme` does install the equivalent of everything that's in `goroot` when you use it to install a go version (and not only the binaries), so I wonder if it has a specific reason to do so.

setup_go.sh
KSerrania: The only thing I could find is old docs from when the `CC_FOR_TARGET` option was introduced, but I believe the nuance is that `CC_FOR_TARGET` is used by `cgo` _when cross-compiling_: https://github.com/golang/go/commit/88d07b2cbcf0ad5e2e9936bcaec4fd25e38bec10, whereas `CC` is used for `cgo` when not cross-compiling (as the `for GOHOSTARCH` / `for GOARCH` distinction seems to show).

setup_go.sh
mx-psi: It looks like the official Go tar.gz for Linux includes the source code and some more stuff (e.g. IANA timezone database info), so it's probably better to keep it. Marking this as solved.

setup_go.sh
```diff
@@ -0,0 +1,23 @@
...
+git checkout go$GO_VERSION
```
mx-psi, line 16:
```suggestion
git clone --branch go$GO_VERSION --depth 1 https://go.googlesource.com/go goroot && cd goroot
```

setup_go.sh
mx-psi: I see. So if I understood correctly, the Go compiler uses cgo and these variables refer to the C compiler that you use for cgo when compiling/cross-compiling the Go compiler, but not to the one used by the Go compiler, which is independent, right?

KSerrania: Yes, I believe that's it. To verify that this is correct, I'll modify the script to remove the `gcc 4.9` compiler after building go.

KSerrania: Yes, I believe that's it. To verify that this is correct, I'll modify the script to remove the `gcc 4.9` compiler after building go.

setup_go.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+set -ex
+
+# Install gimme to get go1.15.11
+curl -sL -o /bin/gimme https://raw.githubusercontent.com/travis-ci/gimme/master/gimme
+chmod +x /bin/gimme
+eval $(gimme 1.15.11)
```
sgnn7, line 8:
```suggestion
eval "$(gimme 1.15.11)"
```

setup_go.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+set -ex
+
+# Install gimme to get go1.15.11
+curl -sL -o /bin/gimme https://raw.githubusercontent.com/travis-ci/gimme/master/gimme
```
sgnn7, line 6: Is there any way to use a stable version with a hash check of this file?

setup_go.sh
```diff
@@ -0,0 +1,32 @@
...
+chmod +x /bin/gimme
+eval $(gimme 1.15.11)
+
+# Install gcc 4.9 (needed to compile go, as it introduces -fdiagnostics-color which is required
+# in the post-compilation steps of go)
+# We don't add it to the PATH so that it's not used in the Agent / rtloader builds
+apt-get update && apt-get install -y gcc-4.9-backport
+
+git clone --branch go$GO_VERSION --depth 1 https://go.googlesource.com/go goroot && cd goroot
```
sgnn7, line 15:
```suggestion
git clone --branch "go$GO_VERSION" --depth 1 https://go.googlesource.com/go goroot && cd goroot
```

setup_go.sh
```diff
@@ -0,0 +1,32 @@
...
+cd src
```
sgnn7, line 16: I'm a bit curious why `cd goroot` is inlined above but we cd again here on a separate line. Probably should join those two together.

setup_go.sh
```diff
@@ -0,0 +1,32 @@
...
+
+# Remove gcc 4.9 after building go
+apt-get remove -y gcc-4.9-backport
+
+# HACK: cgo tries to look for gcc in the same place that CC
+# pointed to when go was compiled. By default, this is fine (since CC=gcc),
+# but here since CC is more specific, cgo will try to find gcc
+# in /usr/lib/gcc-4.9-backport/bin/gcc
+mkdir -p /usr/lib/gcc-4.9-backport/bin
+ln -s $(which gcc) /usr/lib/gcc-4.9-backport/bin/gcc
```
sgnn7, line 29:
```suggestion
ln -s "$(which gcc)" /usr/lib/gcc-4.9-backport/bin/gcc
```

sgnn7 (commented):
Left just a few tiny nit comments

setup_go.sh
KSerrania: I can add a check for this (I copypasted what was in the `Dockerfile` of the builder before).

setup_go.sh
KSerrania: Yeah, it comes from a previous version of this script where the clone + checkout were done in 2 steps.

mx-psi (approved):
LGTM

setup_go.sh
```diff
@@ -0,0 +1,38 @@
...
+
+# Use gcc 4.9 + go1.15.11 to build the target go version
+# HACK: cgo tries to look for gcc in the same place that CC
+# pointed to when go was compiled. By default, CC=gcc.
+# We don't want to keep gcc 4.9 after building go, so we temporarily
+# change the PATH so that "gcc" is the gcc 4.9 we installed.
+# Then, when cgo is used, it will use "gcc" which will be gcc 4.7.
+export PATH="/usr/lib/gcc-4.9-backport/bin:$PATH"
+./all.bash
+
```
djmitche, line 33: It [looks like](https://github.com/golang/go/blob/181e8cde301cd8205489e746334174fee7290c9b/src/make.bash#L212-L213) this "bootstraps" by building the new version with whatever old version it is given (1.15.11 in this case) and then rebuilding the new version with the newly-built binary.  So, it shouldn't matter at all what the old version is, as it only needs to be able to build a working compiler.

djmitche (approved):

PR #162, opened 2021-09-16
Create omnibus-nikos_arm64 image & fix omnibus-nikos_x64 build

ISauve:
- Creates a new build image so we can build `omnibus-nikos` on arm
- Supercedes #159 with a better (less hacky) fix to the openssl issues affecting the `omnibus-nikos` build on x64
- Adds `pkg-config` to the system-probe arm build image

### DIFF
--- a/omnibus-nikos_arm64/Dockerfile
+++ b/omnibus-nikos_arm64/Dockerfile
@@ -1,0 +1,58 @@
+FROM arm64v8/ubuntu:16.04
+
+# Build Args
+ARG GIMME_GO_VERSION=1.16.7
+ARG CLANG_VERSION=8.0.0
+
+# Environment
+ENV GOPATH /go
+ENV CONDA_PATH /root/miniconda3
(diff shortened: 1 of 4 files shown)


omnibus-nikos_arm64/Dockerfile
```diff
@@ -0,0 +1,58 @@
+FROM arm64v8/ubuntu:16.04
```
brycekahle, line 1: This is different from what we use for the system-probe `arm64` build. Do you foresee any problems caused by that?

ISauve: I don't believe so, just because that's also true for the `x64` build! Plus the `build_omnibus-nikos_arm64` job runs on the same gitlab runner as `build_system-probe-arm64`, so if the underlying architecture is the same then I think the important things like word size / endianness / etc. should be the same too, right?

brycekahle (approved):

PR #167, opened 2021-10-01
Update OpenSSL to fix connection errors after Let's Encrypt root cert expiration

albertvaka:
Context:
The system openssl was old and due to a [bug](https://www.openssl.org/blog/blog/2021/09/13/LetsEncryptRootCertExpire/) it was trying to use an expired cert and failing when connecting to sites that used letsencrypt after their root cert changed.

Fix:
- Update `curl` (using a statically-linked binary) to a newer version, which doesn't use the system certs.
- Make `rvm` build ruby to use openssl from miniconda (which is kept up-to-date, together with their its own cacerts) instead of the system openssl.
- Remove `wget` from all builders, which would use the old system openssl

### DIFF
--- a/deb-x64/Dockerfile
+++ b/deb-x64/Dockerfile
@@ -1,3 +1,7 @@
+FROM ubuntu as CURL_GETTER
+RUN apt-get update && apt-get install -y wget
+RUN wget https://github.com/moparisthebest/static-curl/releases/download/v7.79.1/curl-amd64
+
 FROM debian:wheezy-backports
 
 # Build Args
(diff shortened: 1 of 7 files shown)


deb-x64/Dockerfile
```diff
@@ -1,3 +1,7 @@
+FROM ubuntu as CURL_GETTER
```
mx-psi, line 1: We probably want this on ARM too, right?

rpm-x64/Dockerfile
```diff
@@ -37,6 +41,13 @@ RUN yum -y install \
   bzip2-devel e2fsprogs-devel file-devel libacl-devel libattr-devel \
   libxml2-devel lzo-devel nss nss-devel popt-devel sharutils xz-devel
 
+# We install our own ruby, let's remove the system one. It made rvm fail to build ruby for some reason
+RUN yum remove -y ruby
```
mx-psi, line 45: We don't *need* to do this (but maybe it's a good idea to keep it anyway)

deb-x64/Dockerfile
albertvaka: We don't have the problem there, though, since everything is a bit newer... I don't have a strong opinion either way.

rpm-x64/Dockerfile
albertvaka: I decided to keep it, just in case. We never ever should want to use the ruby 1.8 from the system.

deb-x64/Dockerfile
mx-psi: 👍, let's keep it like this. We can clean it up on a future PR if we want to have the same version everywhere

albertvaka: Ended up changing it :P

rpm-x64/Dockerfile
```diff
@@ -37,6 +41,13 @@ RUN yum -y install \
...
+
+# Update curl with a statically linked binary
+COPY --from=CURL_GETTER /curl-amd64 /usr/bin/curl
```
mx-psi, line 48: We may need to remove `curl-devel` from the list above

albertvaka: We need `curl-devel` because we build `git` which needs it. It probably means `yum` is also installing the `curl` binary as a dependency of `curl-devel`, though.

mx-psi (commented):
(Approved)

PR #170, opened 2021-10-04
Upgrade msys then downgrade gcc

albertvaka:
Changes to the Windows builders:
- Upgrade msys to 20210725
- Downgrade gcc to 10.2 due to an incompatibility between gcc 10.3 and go 1.15 and 1.16
- Remove unused x86 builders

Packages for gcc 10.2 are mirrored in the `dd-agent-omnibus` bucket.

### DIFF
--- a/windows/Dockerfile
+++ b/windows/Dockerfile
@@ -35,10 +35,7 @@ ENV RUBY_VERSION "2.6.6-1"
 ENV PYTHON_VERSION "3.8.2"
 ENV WIX_VERSION "3.11.2"
 ENV CMAKE_VERSION "3.17.2"
-# WARN: We currently can't update the msys distribution, as updating it gives us a version of go that's too new,
-# which makes go fail during -race tests. See the discussion in this PR: https://github.com/DataDog/datadog-agent/pull/9074
-# TODO: check when this issue is fixed in Go, then try updating again.
-ENV MSYS_VERSION "20201109.0.0"
+ENV MSYS_VERSION "20210725"
 ENV NUGET_VERSION "5.8.0"
 ENV EMBEDDED_PYTHON_2_VERSION "2.7.17"
 ENV EMBEDDED_PYTHON_3_VERSION "3.8.1"
--- a/windows/Dockerfile
+++ b/windows/Dockerfile
@@ -154,9 +154,10 @@ RUN powershell -C .\install_msys.ps1 -Version $ENV:MSYS_VERSION
 
 RUN ridk install 3
 
-# (32-bit only) Install 32-bit C/C++ compilation toolchain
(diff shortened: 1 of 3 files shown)


windows/Dockerfile
```diff
@@ -157,9 +154,10 @@ RUN powershell -C .\install_msys.ps1 -Version $ENV:MSYS_VERSION
...
-RUN if ($Env:TARGET_ARCH -eq 'x86') { ridk enable; bash -c \"pacman -S --needed --noconfirm mingw-w64-i686-binutils mingw-w64-i686-crt-git mingw-w64-i686-gcc mingw-w64-i686-gcc-libs mingw-w64-i686-headers-git mingw-w64-i686-libmangle-git mingw-w64-i686-libwinpthread-git mingw-w64-i686-make mingw-w64-i686-pkg-config mingw-w64-i686-tools-git mingw-w64-i686-winpthreads-git\" }
-RUN if ($Env:TARGET_ARCH -eq 'x86') { [Environment]::SetEnvironmentVariable(\"Path\", [Environment]::GetEnvironmentVariable(\"Path\", [EnvironmentVariableTarget]::Machine) + \";C:\tools\msys64\mingw32\bin;C:\tools\msys64\usr\bin\", [System.EnvironmentVariableTarget]::Machine) }
+# Downgrade gcc due to an incompatibility between gcc 10.3 and go 1.15 and 1.16
+RUN (New-Object System.Net.WebClient).DownloadFile(\"https://s3.amazonaws.com/dd-agent-omnibus/mingw-w64-x86_64-gcc-10.2.0-11-any.pkg.tar.zst\", \"C:/mingw-w64-x86_64-gcc-10.2.0-11-any.pkg.tar.zst\")
+RUN (New-Object System.Net.WebClient).DownloadFile(\"https://s3.amazonaws.com/dd-agent-omnibus/mingw-w64-x86_64-gcc-libs-10.2.0-11-any.pkg.tar.zst\", \"C:/mingw-w64-x86_64-gcc-libs-10.2.0-11-any.pkg.tar.zst\")
```
julien-lebot, line 159: Shouldn't we also do this for x86 ?

KSerrania: In theory yes, in practice I don't know if it makes sense to spend time fixing builders we don't use (and that we may want to retire in the future)?

julien-lebot: In that case we should remove the x86 jobs from the test stage: `test_windows_1809_x86` && `test_windows_1909_x86`

KSerrania: Yes, either that or re-add the two x86-specific lines for now (to not make the x86 test jobs fail because there's no x86 toolchain available).

albertvaka: Oh, I thought we were not building them anymore (since we are not using them anymore in the Agent pipeline anymore). I'll remove the x86 build and test jobs from here.

julien-lebot (approved):
LGTM and now the pipeline uses slightly less resources thanks for the removal of the x86 jobs ! 😎

PR #181, opened 2021-11-10
Add datadog-ci-uploader image

bkabrda:
In order to install the `datadog-ci` package, necessary to upload JUnit XMLs, NPM is required. Since installing it to other buildimages (which are pretty old and already very big) could be a big hassle, I propose we create this new minimal image that will carry only `datadog-ci` (and Python 3 to run `inv` tasks) and will be used to upload JUnit XMLs produced by jobs that run actual tests.


mx-psi:
Logs say 
```
npm WARN notsup Unsupported engine for @datadog/datadog-ci@0.17.8: wanted: {"node":">=10.24.1"} (current: {"node":"10.19.0","npm":"6.14.4"})
npm WARN notsup Not compatible with your version of node/npm: @datadog/datadog-ci@0.17.8
```

But then it's installed fine. Do we care about that?

bkabrda:
I didn't notice that... I'll take a look, thanks!

bkabrda:
@mx-psi fixed by using https://github.com/tj/n/, please re-review. Thanks!

mx-psi (approved):

PR #184, opened 2021-11-25
Centralize configuration of pip/setuptools versions

bkabrda:
This PR creates a single file that defines used pip/setuptools versions used in all Linux and Windows images. Changing versions in this file will update all our buildimages.


mx-psi (approved):
Looks good. This makes me think that I don't understand why we are keeping the py2 environment here yet. Do you know what do we use it for?

bkabrda:
@mx-psi that's a good question. I don't know actually, I'll try to find out.

bkabrda:
@mx-psi here: https://github.com/DataDog/datadog-agent/blob/main/.gitlab/source_test/linux.yml#L39

mx-psi:
ah, right, for rtloader tests with Python 2. Makes sense, thanks!

PR #194, opened 2022-01-20
[CWS] Install gcc-multilib in system-probe* images

paulcacheux:
The CWS functional tests use a x86 (32 bits) binary in the CI to test x86 specific syscalls and ways of doing syscalls. In order to build those binaries in the CI, we would need `gcc-multilib` installed in the build image.

### DIFF
--- a/system-probe_arm64/Dockerfile
+++ b/system-probe_arm64/Dockerfile
@@ -14,7 +14,6 @@ RUN apt-get update && apt-get install -y --no-install-recommends apt-utils && ap
         flex \
         g++ \
         gcc \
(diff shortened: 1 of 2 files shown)


brycekahle:
> In order to build x86 binaries in the CI

Can you clarify what you mean? Do you mean x86 (32 bit) rather than x86_64?

paulcacheux:
Yes ! We currently build them locally ahead of time but we would like to build them on the CI

brycekahle (approved):

mx-psi (approved):

PR #195, opened 2022-01-25
Build vcpkg-tool from source

julien-lebot:
This PR builds VCPKG from source, and uses it in place of the pre-built VCPKG that is downloaded from https://github.com/microsoft/vcpkg-tool/releases.

The rationale is that we use a `vcpkg` baseline for our dependencies that is not necessarily the latest, so  the `bootstrap.ps1` file that is used to install `vcpkg.exe` in the buildimage, installs an older version of `vcpkg` too.
See : https://github.com/microsoft/vcpkg/blob/5568f110b509a9fd90711978a7cb76bae75bb092/scripts/bootstrap.ps1#L48

By building it ourselves we can always have the latest `vcpkg.exe` in our buildimage.

### DIFF
--- a/windows/Dockerfile
+++ b/windows/Dockerfile
@@ -117,10 +117,6 @@ RUN powershell -Command .\install_dotnetcore.ps1
 COPY ./windows/install_nuget.ps1 install_nuget.ps1
 RUN powershell -Command .\install_nuget.ps1 $ENV:NUGET_VERSION
 
-# install vcpkg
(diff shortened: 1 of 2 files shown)


albertvaka:
Would it work to download a newer binary version without using  `bootstrap.ps1`? Or we would then hit the mysterious `curl` issue? If all we need is a recent version, we can still use the binary releases.

albertvaka (approved):

PR #197, opened 2022-01-31
[all] Install requirements.txt on datadog-agent-buildimages

mx-psi:
Install `requirements.txt` needed to build the Datadog Agent at image build time. After this we need to:
- Do the same thing on datadog-agent-builders
- Point the datadog-agent `requirements.txt` file to the `requirements.txt` file here
- (possibly) Remove the `pip install requirements.txt` lines from the Gitlab CI
- Move Dependabot here

### DIFF
--- a/setup_python.sh
+++ b/setup_python.sh
@@ -77,7 +77,7 @@ conda create -n ddpy3 python python=$PY3_VERSION
 conda activate ddpy2
 pip install -i https://pypi.python.org/simple pip==${DD_PIP_VERSION}
 pip install setuptools==${DD_SETUPTOOLS_VERSION}
-pip install invoke==1.4.1 distro==1.4.0 awscli==1.16.240
+pip install distro==1.4.0 awscli==1.16.240
 
 # Update pip, setuptools and misc deps
 conda activate ddpy3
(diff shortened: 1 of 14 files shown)


brycekahle (approved):

windows/install_embedded_pythons.ps1
```diff
@@ -78,8 +78,9 @@ curl https://bootstrap.pypa.io/pip/2.7/get-pip.py -o get-pip.py
 cd $py3Target
 curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py
 .\python get-pip.py pip==${Env:DD_PIP_VERSION_PY3}
+.\python -m pip install -r requirements.txt
```
julien-lebot, line 81: nit: I would put this one line lower to ensure it doesn't get confused as being part of the py3 block.

setup_python.sh
```diff
@@ -76,13 +77,15 @@ conda create -n ddpy3 python python=$PY3_VERSION
 conda activate ddpy2
 pip install -i https://pypi.python.org/simple pip==${DD_PIP_VERSION}
 pip install setuptools==${DD_SETUPTOOLS_VERSION}
-pip install distro==1.4.0 awscli==1.16.240
+pip install invoke==1.4.1 distro==1.4.0 awscli==1.16.240
```
julien-lebot, line 80: why are we installing `invoke==1.4.1` when `requirements.txt` specify `1.6.0` ?

mx-psi: Removed, no need to change this

windows/install_embedded_pythons.ps1
mx-psi: This **is** part of the Python 3 block; requirements.txt can't even be installed on Python 2 (some dependencies are incompatible)

julien-lebot (approved):
LGTM

PR #198, opened 2022-01-31
[deb-armhf] build Python 3.9.9 from source

mx-psi:
Due to deadsnakes/issues#195 we need to find an alternative source for installing Python on the deb-armhf builder


albertvaka (commented):
I think this should be in `setup_python.sh` for consistency given every image installs python using that script.

mx-psi:
I tried to do this on a separate stage (thus the different Dockerfile), but I haven't managed to troubleshoot the issue I was having with it, so I am going to move it to `setup_python.sh` 👍

albertvaka (approved):
If the reason to build Python in a separate stage was to avoid creating intermediate images containing the sources and build intermediate objects, you can still delete those files form `setup_python.sh` and they won't be in any intermediate image (since the whole script runs in a single build step). Other than that, I see little advantage in doing multi-stage.

setup_python.sh
```diff
@@ -26,23 +26,36 @@ case $DD_TARGET_ARCH in
...
+        zip wget build-essential checkinstall libreadline-gplv2-dev \
+        libncursesw5-dev libssl-dev libsqlite3-dev tk-dev libgdbm-dev \
+        libc6-dev libbz2-dev libffi-dev zlib1g-dev
+        wget https://www.python.org/ftp/python/3.9.9/Python-3.9.9.tgz
+        tar xzf Python-3.9.9.tgz
+        pushd /Python-3.9.9
+           ./configure
+           make -j 8
+           make install
+        popd
```
julien-lebot, line 44: We don't need the build source & build directory after make install.
```suggestion
        popd
        rm -rf Python-3.9.9
```

julien-lebot (changes requested):
Left a comment for cleaning up after building Python in container.

setup_python.sh
julien-lebot: @mx-psi we also need to remove the tarball archive.

mx-psi: ✅

julien-lebot (approved):

PR #199, opened 2022-01-31
[rpm-armhf] Replace get-pip.py URL

mx-psi:
The `python3-devel` package installs Python 3.6, and we need an older `get-pip.py` script for this Python version. Relevant job output: https://gitlab.ddbuild.io/DataDog/datadog-agent-buildimages/-/jobs/112687251

### DIFF
--- a/setup_python.sh
+++ b/setup_python.sh
@@ -36,7 +36,7 @@ case $DD_TARGET_ARCH in
         ln -sf /usr/bin/python3.9 /usr/bin/python3
         DD_GET_PIP_URL=https://bootstrap.pypa.io/pip/get-pip.py
     elif [ -f /etc/redhat-release ] || [ "$DISTRIBUTION" == "RedHat" ] || [ "$DISTRIBUTION" == "CentOS" ] || [ "$DISTRIBUTION" == "Amazon" ]; then
-        yum install -y python3-devel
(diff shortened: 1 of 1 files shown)


setup_python.sh
```diff
@@ -34,11 +34,13 @@ case $DD_TARGET_ARCH in
         add-apt-repository -y ppa:deadsnakes/ppa
         apt-get update && apt-get install -y python3.9-dev python3.9-distutils
         ln -sf /usr/bin/python3.9 /usr/bin/python3
+        DD_GET_PIP_URL=https://bootstrap.pypa.io/pip/get-pip.py
     elif [ -f /etc/redhat-release ] || [ "$DISTRIBUTION" == "RedHat" ] || [ "$DISTRIBUTION" == "CentOS" ] || [ "$DISTRIBUTION" == "Amazon" ]; then
         yum install -y python3-devel
```
albertvaka, line 39:
```suggestion
        yum install -y python3-devel # This installs python 3.6 on arm32v7/centos:7
```

albertvaka (approved):
LGTM. deb_armhf are still failing because python 3.9 is not available, but we can fix that on a separate PR.

mx-psi:
> deb_armhf are still failing because python 3.9 is not available, but we can fix that on a separate PR.

That would be #198 :)

PR #208, opened 2022-02-24
Add dd-agent-testing image

albertvaka:
Makes public the image currently used to trigger kitchen tests.


albertvaka:
TODO: Add checksums

PR #209, opened 2022-02-25
Verify sha256 checksums of all downloaded artifacts

KSerrania:
### What does this PR do?

Adds sha256 checksum verifications for all artifacts we download, and use HTTPS whenever possible.
Pins all artifacts that weren't previously pinned.
Checksums have been first computed locally then verified in the CI.

### Motivation

Check that the components we're using haven't been tampered with. More reproducible builds.

### Additional notes

Future improvements:
- on Linux, we download artifacts through various ways (`wget`, `curl` with different option sets). We should check whether or not we can harmonize this and have a single (or a few) utility functions to download & check the sums of artifacts.
- on Windows, some install scripts define utility functions (`DownloadFile` and `DownloadAndExpandTo`) that could likely be reused in other install scripts, and which could include the checksum verification.

### DIFF
--- a/windows/install_python.ps1
+++ b/windows/install_python.ps1
@@ -1,5 +1,6 @@
 param (
-    [Parameter(Mandatory=$true)][string]$Version
+    [Parameter(Mandatory=$true)][string]$Version,
+    [Parameter(Mandatory=$true)][string]$Sha256
 )
 
 # Enabled TLS12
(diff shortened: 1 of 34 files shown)


brycekahle (approved):

KSerrania:
Given the failed Windows builds, it seems like the AWS CLI package we fetch isn't pinned, and therefore may change at any time. If we want to check its sum, we'll have to find another location where we can get a specific version of the AWS CLI.

albertvaka:
I wonder why we need to install it from an MSI on Windows while we can install it using pip on Linux. Maybe it's enough to install with pip?

KSerrania:
I'm currently trying that in a test branch. We can install awscli with pip, but that will require a `datadog-agent` change, to stop calling `aws.exe` (which isn't installed with the pip package) and use `aws` instead.

windows/install_python.ps1
```diff
@@ -27,4 +27,8 @@ setx PATH "$($Env:PATH);c:\program files\Python38;c:\Program files\python38\scri
 $Env:PATH="$($Env:PATH);c:\program files\Python38;c:\Program files\python38\scripts"
 Remove-Item $out
 
+curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py
+python get-pip.py pip==${Env:DD_PIP_VERSION_PY3}
+python -m pip install awscli==1.22.57
```
mx-psi, line 32: This is already present in requirements.txt, isn't it?

mx-psi (approved):

PR #210, opened 2022-02-25
python deps: Remove distro, add pyyaml on py2

albertvaka:


windows/Dockerfile
```diff
@@ -142,7 +142,7 @@ RUN setx GOPATH C:\dev\go
 # Install system Python 3 (to use invoke).
 # We always install the 64 bit version because vcredist140 won't work otherwise
 COPY ./windows/install_python.ps1 install_python.ps1
-COPY ./requirements.txt /requirements.txt
+COPY ./requirements.txt ./requirements-py2.txt /
```
KSerrania, line 145: The new `requirements-py2.txt` file isn't used yet, I believe you need to use it here: https://github.com/DataDog/datadog-agent-buildimages/blob/29f5a0b56e0452efbd51c16c97436be052d3b53b/windows/install_embedded_pythons.ps1#L73-L76

(btw, the line to install the Python 3 deps in that same file is wrong: https://github.com/DataDog/datadog-agent-buildimages/blob/29f5a0b56e0452efbd51c16c97436be052d3b53b/windows/install_embedded_pythons.ps1#L81 `requirements.txt` isn't present there as this command is run within the `$py3Target` directory)

albertvaka: I'm now installing `requirements-py2.txt`, I'm calling python without changing the working dir so it finds `requirements.txt` and I've added a check so the script fails if those commands fail. Let's see if any of my changes work.

albertvaka:
Test pipeline: https://gitlab.ddbuild.io/DataDog/datadog-agent/-/pipelines/7147427

KSerrania (approved):

PR #217, opened 2022-03-15
Update deb_x64 builder to Debian Jessie

bkabrda:
As noted in the omnibus-nikos Dockerfile, we can't update that one, as that would break compatibility with CentOS 7.

### DIFF
--- a/deb-x64/Dockerfile
+++ b/deb-x64/Dockerfile
@@ -2,7 +2,7 @@ FROM ubuntu as CURL_GETTER
 RUN apt-get update && apt-get install -y wget
 RUN wget https://github.com/moparisthebest/static-curl/releases/download/v7.79.1/curl-amd64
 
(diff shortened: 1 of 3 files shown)


ISauve (approved):

deb-x64/Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM ubuntu as CURL_GETTER
 RUN apt-get update && apt-get install -y wget
 RUN wget https://github.com/moparisthebest/static-curl/releases/download/v7.79.1/curl-amd64
 
-FROM debian:wheezy-backports
+FROM debian:jessie-backports
```
mx-psi, line 5: Should we keep building both the Wheezy and Jessie versions until we release 7.36? We might potentially have to rebuild the Debian Wheezy image if we have to do a 7.34.x release, right?

PR #220, opened 2022-03-31
`pip install requirements.txt` also for Python 2

juliogreff:
This installs the Python dependencies for both versions (2 and 3),
instead of requiring the `datadog-agent` jobs running py2 to also run
`pip install requirements.txt`. This will allow us to remove this
install step entirely from `datadog-agent` (see
https://github.com/DataDog/datadog-agent/pull/11511)


setup_python.sh
```diff
@@ -84,6 +84,7 @@ conda activate ddpy2
 pip install -i https://pypi.python.org/simple pip==${DD_PIP_VERSION}
 pip install setuptools==${DD_SETUPTOOLS_VERSION}
 pip install -r requirements-py2.txt
+python3 -m pip install -r requirements.txt
```
mx-psi, line 87: I would rather move this to line 68, just after the case statement, since it is not related to Python 2 in any way

mx-psi (commented):
We use the system Python 3 for invoke tasks, and the ddpy2 Python 2 environment is only used within the rtloader tests, it should never be used for tooling. We are probably missing a requirements.txt install somewhere, but I am not sure that this is the right place.

What Gitlab jobs were failing and how where they failing? That will help me understand where this should be put

mx-psi:
As discussed on Slack: the current status is that `-py3` jobs use `ddpy3`-invoke while `-py2` jobs use the "system Python 3"-invoke. This solution makes sense, I just want https://github.com/DataDog/datadog-agent-buildimages/pull/220#discussion_r839658397 for clarity

setup_python.sh
```diff
@@ -75,6 +77,10 @@ PATH="${CONDA_PATH}/bin:${PATH}"
 conda init bash
 source /root/.bashrc
 
+which python
+which python3
+python3 -m pip install -r requirements.txt
```
albertvaka, line 82: Let's add a comment here saying that this is for the system python.

KSerrania:
> An alternative would be to use the py3 conda's python binary explicitly to run invoke, while having the py2 conda active. That would avoid using the system python altogether, which is good.

Yes, that'd be another option to stop using the system python (which we don't really manage). I'm trying it out, I'll report here if that works.

PR #221, opened 2022-04-01
Create base `docker` image with python dependencies installed

juliogreff:
This prevents a `pip install` on every `datadog-agent` image build,
making builds a little bit faster and more reliable.

### DIFF
--- a/docker-arm64/Dockerfile
+++ b/docker-arm64/Dockerfile
@@ -1,0 +1,5 @@
+FROM 486234852809.dkr.ecr.us-east-1.amazonaws.com/docker:v2718787-3888eda-18.09.6-arm64-py3
+
+COPY ./requirements.txt .
+
+RUN python3 -m pip install -r requirements.txt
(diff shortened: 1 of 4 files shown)


albertvaka (approved):

docker-arm64/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM 486234852809.dkr.ecr.us-east-1.amazonaws.com/docker:v2718787-3888eda-18.09.6-arm64-py3
```
julien-lebot, line 1: There is no equivalent public image for this, so our customer won't be able to use it

juliogreff: That is already the case with several of the images in the `datadog-agent` pipelines though, no? We're just moving them here instead of there

julien-lebot: Not in this repository, which is our public-facing build images.

albertvaka: I think it's fine to have them here for simplicity, although it's true that until now all images here were usable by external people and this one is not.

PR #227, opened 2022-04-08
Upgrade CMake to 3.23.0

albertvaka:
Bumps CMake cross all buildimages to 3.23.0 (latest).

Before we were installing 3.14.4 on Linux, 3.17.2 on Windows and the one from the distro repos on the nikos buildimages.

For aarch64, we can now use upstream binaries instead of building it ourselves.


paulcacheux (approved):

omnibus-nikos_arm64/Dockerfile
```diff
@@ -13,23 +13,36 @@ FROM arm64v8/ubuntu:16.04
...
 # Environment
 ENV GOPATH /go
 ENV GIMME_GO_VERSION $GIMME_GO_VERSION
+ENV CMAKE_VERSION $CMAKE_VERSION
+ENV CMAKE_SHA256 $CMAKE_SHA256
 ENV CLANG_VERSION $CLANG_VERSION
 ENV CLANG_SHA256 $CLANG_SHA256
 
-RUN apt-get update && apt-get install -y fakeroot cmake curl bzip2 g++ gcc git \
+RUN apt-get update && apt-get install -y fakeroot procps bzip2 g++ gcc git \
```
brycekahle, line 29: `curl` is used below, so should still be required, yes?

albertvaka: The first lines of the Dockerfile already install a specific, pinned version of curl, so we don't depend on the system one. This one shouldn't be needed.

albertvaka:
Test pipeline with cmake fix: https://gitlab.ddbuild.io/DataDog/datadog-agent-buildimages/-/pipelines/7629862

omnibus-nikos_arm64/Dockerfile
brycekahle: 👍

julien-lebot (approved):

PR #228, opened 2022-04-14
Regenerate and update "dockerized" powershell script.

iglendd:
The script "dockerfile-to-powershell-generated-sample.ps1" can/should be
used to make Windows build-like machine from a "plain" Windows machine.

This needs to be done every time Windows/Dockerfile changes (until we automate the process),  

The script had been (re)generated via
    dockerfile-to-powershell.py
        -d .\Dockerfile
        -p .\dockerfile-to-powershell-generated-sample.ps1
        -a WINDOWS_VERSION=1809
        -a DD_TARGET_ARCH=x64


windows/dockerfile-to-powershell-generated-sample.ps1
```diff
@@ -97,113 +97,146 @@ function Update-SessionEnvironment {
...
 
-# Line: 62, Content: LABEL nuget_version=${NUGET_VERSION}
+# Line: 73, Content: LABEL nuget_version=${NUGET_VERSION}
 # LABEL resolved: nuget_version='5.8.0'
 
 # Add certificates needed for build & check certificates file hash
 # We need to trust the DigiCert High Assurance EV Root CA certificate, which signs python.org,
 # to be able to download some Python components during the Agent build.
-# Line: 66, Content: RUN (New-Object System.Net.WebClient).DownloadFile(\"https://curl.haxx.se/ca/cacert-${ENV:CACERTS_VERSION}.pem\", \"cacert.pem\")
+# Line: 77, Content: RUN (New-Object System.Net.WebClient).DownloadFile(\"https://curl.haxx.se/ca/cacert-${ENV:CACERTS_VERSION}.pem\", \"cacert.pem\")
```
julien-lebot, line 236: Not including the line number could help save a bit on the diff count ?

windows/dockerfile-to-powershell-generated-sample.ps1
```diff
@@ -97,113 +97,146 @@ function Update-SessionEnvironment {
...
 $stopwatch = [system.diagnostics.stopwatch]::StartNew()
 Write-Host @'
-[1/59] Starting to (New-Object System.Net.WebClient).DownloadFile("https://curl.haxx.se/ca/cacert-${ENV:CACERTS_VERSION}.pem", "cacert.pem")
+[1/57] Starting to (New-Object System.Net.WebClient).DownloadFile("https://curl.haxx.se/ca/cacert-${ENV:CACERTS_VERSION}.pem", "cacert.pem")
```
julien-lebot, line 239: Same for those I'm presuming steps number ? They're probably useful for debugging, but we could maybe put them behind a feature flag so that the final script doesn't contain them ? It would help cut down the number of diff when updating the script.

julien-lebot (approved):
✅  with some comments - there's a lot of diff that's just changing line number/step count.

windows/dockerfile-to-powershell-generated-sample.ps1
iglendd: Technically code-review here is a formality because the file is fully generated. Of course you are right but you it helps a lot to pin-point a problem and trace it back to the docker file.

windows/dockerfile-to-powershell-generated-sample.ps1
iglendd: The same argument as above. We are actually hope we will automate it. In this case it is not for debugging but to see in a very long script where are you. Also if a script fails I can much easier to find a place from which to restart.

windows/dockerfile-to-powershell-generated-sample.ps1
julien-lebot: I think the line content is enough to know where it failed, there are not a lot of similar lines in this file.

iglendd:
Per team's discussion
* Keep this PR AS IS
* Add another to remove mentioned by @julien-lebot variability by default and have additional argument (e.g. debug) to turn them on.

Issue #235, opened 2022-04-27
Package manager cleanning


katepangLiu:
- yum
  - all yum commands should end with " && yum clean all"
- apt-get
  - Official Debian and Ubuntu images [automatically run apt-get clean](https://github.com/moby/moby/blob/03e2923e42446dbb830c654d0eec323a0b4ef02a/contrib/mkimage/debootstrap#L82-L105), so explicit invocation is not required.

PR #236, opened 2022-04-27
Add yum clean

katepangLiu:
#235


albertvaka (approved):
Thanks for the PR! Looks good to me ✅ 

I didn't know about `apt` doing this by default and I found it quite curious 😮, thanks for linking that code on the issue as well 😃

julien-lebot (approved):

PR #238, opened 2022-04-27
Use Windows line endings for the powershell generator even on Linux

bkabrda:
Without this PR, running the generator on Linux transforms all the newlines to just `\n` which makes the whole file appear as a git diff.

### DIFF
--- a/windows/dockerfile-to-powershell.py
+++ b/windows/dockerfile-to-powershell.py
@@ -2,6 +2,7 @@ from ast import Not
 import sys
 import getopt
 import re
+import io
 from datetime import datetime
 from dockerfile_parse import DockerfileParser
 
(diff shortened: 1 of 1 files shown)


windows/dockerfile-to-powershell.py
```diff
@@ -171,7 +172,7 @@ def parseDockerFile(dockerFilePath):
 # Open/Create Powershell file for writing
 def createPowershellFile(powershellFilePath):
    try:
-      return open(powershellFilePath, 'w')
+      return io.open(powershellFilePath, 'w', newline="\r\n")
```
julien-lebot, line 175: Is there a cross-platform way of doing that ?

iglendd: ... it is cross-compiling way to work

iglendd (approved):

windows/dockerfile-to-powershell.py
bkabrda: There is none AFAICS :/

PR #240, opened 2022-05-02
[cacerts] 2022-04-26 update

KSerrania:


albertvaka (approved):

julien-lebot (approved):

iglendd (commented):
what about docerizing script :(

KSerrania:
> what about docerizing script :(

I didn't run the dockerize script, but I did apply the changes to the `windows/dockerfile-to-powershell-generated-sample.ps1` file (through automatic search / replace), which AFAIK achieves the same result in this case (except for the date in the header of the file), doesn't it? I'll run it and commit the result in case that changes something, but I don't think it will.

> We should find a way to run that script automatically whenever the build image changes. Could be a pre-commit step, wdyt @iglendd ?

Yes, we already have a card in our backlog to tackle that.

iglendd:
> We should find a way to run that script automatically whenever the build image changes. Could be a pre-commit step, wdyt
That is probably better idea than https://datadoghq.atlassian.net/browse/AP-1558. What do you think @albertvaka. I personally did not expect that Dockerfile would be changing that often

iglendd (approved):

PR #245, opened 2022-05-17
Fix call to pip install

juliogreff:
Missing a `-r`, fixes the following error, while also actually installing python deps: `ERROR: Invalid requirement: '../requirements.txt'`


amenasria (approved):
LGTM !

PR #246, opened 2022-05-25
[system-probe] Install newer kernel headers using buster-backports

lebauce:


system-probe_x64/Dockerfile
```diff
@@ -25,6 +28,8 @@ RUN apt-get update && apt-get install -y --no-install-recommends apt-utils && ap
         libxml2-dev \
         libzip-dev \
         linux-headers-amd64 \
+        linux-headers-5.10.0-0.bpo.12-amd64 \
```
brycekahle, line 31: Since the docker image isn't pinned a specific version of `buster`, won't this specific version become outdated?

lebauce: It may indeed be the case. I could not find a reliable way to get the latest kernel headers. Maybe using somethink like linux-headers-5*-amd64 would be better ?

brycekahle:
@lebauce It looks like there is some strange [compile error](https://gitlab.ddbuild.io/DataDog/datadog-agent-buildimages/-/jobs/140588028) with the newer headers:
```
In file included from ./pkg/network/ebpf/c/prebuilt/offset-guess.c:6:
In file included from /usr/src/linux-headers-5.10.0-0.bpo.9-common/include/net/sock.h:59:
In file included from /usr/src/linux-headers-5.10.0-0.bpo.9-common/include/linux/filter.h:19:
/usr/src/linux-headers-5.10.0-0.bpo.9-common/include/linux/set_memory.h:22:19: error: static declaration of 'set_direct_map_default_noflush' follows non-static declaration
static inline int set_direct_map_default_noflush(struct page *page)
                  ^
/usr/src/linux-headers-5.10.0-0.bpo.9-common/arch/x86/include/asm/set_memory.h:84:5: note: previous declaration is here
int set_direct_map_default_noflush(struct page *page);
```

lebauce:
@brycekahle I found why : I installed multiple versions of the headers, it randomly chose 5.10.0-0.bpo.9 and this version seems to be buggy. I now only install the latest kernel headers (5.10.0-0.bpo.12) and this version works.

brycekahle:
@lebauce looks like things are failing for a different reason now. I think @ISauve was looking into it.

lebauce:
@brycekahle `dd-gitlab/test_rpm_x64` and `dd-gitlab/build_omnibus-nikos_x64` seem indeed to be failing on master.

KSerrania (approved):
The two failures are unrelated to this PR, LGTM

brycekahle (approved):

PR #252, opened 2022-06-08
[python] Pin ddpy3 environment deps to prevent Python crashes in rpm image

KSerrania:
### What does this PR do?

Pins a bunch of dependencies installed in the `ddpy3` environment to their last working version.

### Motivation

The `ddpy3` environment is currently broken in the `rpm_x64` image due to some dependencies being updated last week, causing all uses of Python to break with: `libgcc_s.so.1 must be installed for pthread_cancel to work`. This prevents any image currently built from being usable in the `datadog-agent` pipeline.


mx-psi (commented):
There still seems to be an error on the nikos image 🤔

mx-psi:
Oh, I got that wrong then. Makes sense.

mx-psi (approved):

PR #262, opened 2022-07-29
add zstd to build images

paulcacheux:
The end goal of this PR is to compress go deps archives with zstd instead of gzip, to do this we need zstd in deb_x64 for the compression part, and in all images that will unarchive the dependencies.


brycekahle (approved):

bkabrda:
Hi :wave: thanks for the PR. Two questions:
* Could you please create a testing Agent pipeline with the updated buildimages and zstd-compressed deps archive? I'm especially interested in seeing how this works on Windows.
* Could you explain why you want to switch and what implications it will have on pipeline speed on Linux and Windows?

Thanks!

paulcacheux:
I'm still working on understanding the windows story, but here is a PR I made months ago showing improvements, I will update it once I get windows working

windows/helpers/phase1/install_7zip.ps1
```diff
@@ -18,18 +19,17 @@ if($isInstalled -and $isCurrent){
...
 Get-RemoteFile -RemoteFile $sevenzip -LocalFile $out -VerifyHash $Sha256
 
 Start-Process $out -ArgumentList '/S' -Wait
 Remove-Item $out
-Add-ToPath -NewPath "c:\program files\7-zip" -Local -Global
+Add-ToPath -NewPath "c:\program files\7-zip-Zstandard" -Local -Global
 
 ## Write the version key out to the registry
-Set-InstalledVersionKey -Component "sevenzip" -Keyname "Version" -TargetValue $Version
+Set-InstalledVersionKey -Component "sevenzip" -Keyname "version" -TargetValue $Version
```
clarkb7, line 33: is it possible for `$ZstdVersion` to change but `$Version` to remain the same? if so we should combine them for the `Get-InstalledUpgradeStatus` and `Set-InstalledVersionKey` commands.

paulcacheux: Yes I think it's possible, let me update as you said

clarkb7 (approved):

PR #266, opened 2022-08-06
docker image size reduction

paulcacheux:
The current `deb-x64` docker image is around ~7.5GB. Even though it is cached in most cases, when it's not it can take up to 1min30secondes to setup the docker executor in the agent CI because of pull time.

This PR reduces the image size by:
- doing the git installation in one layer (before the installation was split into multiple layers with one containing the downloaded archive)
- installing rust with minimal profile (dropping `rust-docs` and `rustfmt` that are not needed for a CI installation)
- ~cleaning up the apt lists after installation~
- removing the gimme go versions after usage (currently we use gimme to download an old go version, in order to build the final go version from sources, this PR removes the old go version after use)

to arrive at ~6.22GB for the `deb-x64` image

Agent-side pipeline to show that it still passes as before: https://gitlab.ddbuild.io/DataDog/datadog-agent/-/pipelines/9192158

This PR also adds steps in the build script in order to display the image size, but also the size of each layer to enable more introspection in the future.

**For the future**: the current worst offender in the `deb-x64` case is the installation of clang+llvm (the layer takes ~2GB). This installation is also way too big for what we use (it includes lldb for example but also support for a lot of architectures and programming languages that we don't use).

### DIFF
--- a/setup_go.sh
+++ b/setup_go.sh
@@ -20,3 +20,7 @@ git clone --branch "go$GO_VERSION" --depth 1 https://go.googlesource.com/go goro
 
 # Update PATH to include the built go binaries
 echo 'export PATH="/goroot/bin:$PATH"' >> /root/.bashrc
+
+# Remove gimme
+rm -rf $HOME/.gimme
+rm /bin/gimme
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -45,7 +45,7 @@ variables:
     - docker push 486234852809.dkr.ecr.us-east-1.amazonaws.com/ci/datadog-agent-buildimages/$IMAGE:v$CI_PIPELINE_ID-${CI_COMMIT_SHA:0:7}
     # For size debug purposes
     - docker images 486234852809.dkr.ecr.us-east-1.amazonaws.com/ci/datadog-agent-buildimages/$IMAGE:v$CI_PIPELINE_ID-${CI_COMMIT_SHA:0:7}
-    - docker history 486234852809.dkr.ecr.us-east-1.amazonaws.com/ci/datadog-agent-buildimages/$IMAGE:v$CI_PIPELINE_ID-${CI_COMMIT_SHA:0:7}
+    - docker history --no-trunc 486234852809.dkr.ecr.us-east-1.amazonaws.com/ci/datadog-agent-buildimages/$IMAGE:v$CI_PIPELINE_ID-${CI_COMMIT_SHA:0:7}
     # For testing purposes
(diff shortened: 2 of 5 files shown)


brycekahle (approved):

sgnn7:
Just a drive-by with two tiny notes here:
- Removal of lists impacts the duration of these builds if subsequently `apt-get` is used since it has to redownload the state of repos but that may not be something we care about overall.
- `apt-get clean` may be beneficial in all of these install commands since the package cache will probably be even a bigger space hog.
- `apt-get autoremove` may also be beneficial to clean a few things up but it may have unintended effects depending on if anything gets removed that we implicitly depend on.

paulcacheux:
To be fair, the apt stuff is not the biggest improvement of this PR, I'm fine with removing it if desired. On the other hand I thought the entire basis of this repo was to not do any apt on the agent/CI side

paulcacheux:
The apt clean step should not be beneficial since it should already be done automatically: https://github.com/moby/moby/blob/03e2923e42446dbb830c654d0eec323a0b4ef02a/contrib/mkimage/debootstrap#L82-L105

In the docs: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/
```
Official Debian and Ubuntu images [automatically run apt-get clean](https://github.com/moby/moby/blob/03e2923e42446dbb830c654d0eec323a0b4ef02a/contrib/mkimage/debootstrap#L82-L105), so explicit invocation is not required
```

paulcacheux:
Alright I remove the apt list cleanup, we are now at `6.67GB` still a win, if we want we can reintroduce this cleanup later

setup_go.sh
```diff
@@ -20,3 +20,7 @@ git clone --branch "go$GO_VERSION" --depth 1 https://go.googlesource.com/go goro
 
 # Update PATH to include the built go binaries
 echo 'export PATH="/goroot/bin:$PATH"' >> /root/.bashrc
+
+# Remove gimme
+rm -rf $HOME/.gimme
+rm /bin/gimme
```
mx-psi, line 26: `gimme` is [used in the security agent build](https://github.com/DataDog/datadog-agent/blob/e57ba47144922e7c779d55d13083aead7edc6b94/tasks/security_agent.py#L45), but I am not sure why removing it here did not make the tests fail :thinking:

paulcacheux: It's used there only if we want to override the go version, something we never do in the CI and that to my knowledge no one ever does locally. I will do a PR to remove this thing if unused

mx-psi: Ah, right, didn't notice the `if go_version`.

mx-psi (approved):

setup_go.sh
paulcacheux: This should be good now, the use in security agent was removed

PR #269, opened 2022-08-17
Adding codeql to build image for code scanning during CI

derekwbrown:
Title says it all.

Add the github/codeql software to build image and path to allow generating codeql databases during ci.


windows/install_codeql.ps1
```diff
@@ -0,0 +1,24 @@
...
+Start-Process "7z" -ArgumentList "x -oc:\ $out" -Wait
+
+Write-Host -ForegroundColor Green "Removing temporary file $out"
+
+Remove-Item $out
+
+setx PATH "$Env:Path;c:\CodeQL;"
+$Env:Path="$Env:Path;c:\CodeQL;"
+
+Write-Host -ForegroundColor Green "Installed go $ENV:GO_VERSION"
```
KSerrania, line 24:
```suggestion
Write-Host -ForegroundColor Green "Installed CodeQL $ENV:CODEQL_VERSION"
```

windows/install_codeql.ps1
```diff
@@ -0,0 +1,24 @@
...
+$ProgressPreference = 'SilentlyContinue'
+
+Write-Host -ForegroundColor Green "Installing CodeQL $ENV:CODEQL_VERSION"
+#             https://github.com/github/codeql-cli-binaries/releases/download/v2.10.3/codeql-win64.zip
+$codeqlzip = "https://github.com/github/codeql-cli-binaries/releases/download/v$ENV:CODEQL_VERSION/codeql-win64.zip"
+$out = 'codeql.zip'
+
+Write-Host -ForegroundColor Green "Downloading $codeqlzip to $out"
+
+(New-Object System.Net.WebClient).DownloadFile($codeqlzip, $out)
```
KSerrania, line 11: Could you add a hash check, same as what we do for other files we download, eg. [here](https://github.com/DataDog/datadog-agent-buildimages/blob/746a71dcf378fb633efb07ffcefb35d510b41d45/windows/install_go.ps1#L13)?

derekwbrown: AFAICT; since it's a github release download, there's no published hash.  
I could download and hash it myself, but I don't see what security that would add. 
Since the download URL is `https` and github is trusted, that should be sufficient, yes?

KSerrania: Agreed on your points. The part that I'd say makes it useful is to make sure what we download doesn't change over time without us knowing - we want to know if what is behind a given a certain URL changes, even if we trust the domain's owner.

KSerrania (approved):

PR #276, opened 2022-09-05
Refactor tools to allow building container and setting up dev machine

derekwbrown:
# Objective
This PR attempts to solve the following problems

1.  Maintain the build container with little change to adding/modifying software
2.  Create a local build environment using the same scripts as the build container, so that the local build environment matches the container build environment
3.  Allow the local build environment to be run multiple times, to update/add new tools as necessary

# Description
This PR modifies the helper scripts for the container building into a single top-level build script.  This allows that same build script to be run on a local machine, ensuring the local build tools match the container build tools.

The script also saves the current state of the tools (what version has been installed), so that it can be re run. When it is run for the N+1th time, it checks the version installed and skips software that's already correct (so no multiple installs of vstudio).  For versions that have changed, it will upgrade to the new version.  For tools that have been added since the last run, it will add the new tools.

# Using the local build environment

When running in local mode (not when building the container), the script refrains from changing the system environment wherever possible.  It does not add environment variables and change the path.

The local build mode includes a powershell module (`DDDeveloper`), which includes powershell commands to enable the environment. Once installed, the developer can execute `Use-BuildEnv` and the module will update the environment (path and necessary variables) for _that shell only_, which avoids polluting the general setup.

## A Note on GO
In the local environment, this tool allows multiple concurrent versions of go.  This is to support wanting to maintain > 1 older versions of go in the event that switching is necessary between versions.  When running `Use-BuildEnv`, by default, it will select the most recent version of Go installed on the machine.  However, it accepts an argument `-GoVer` in which the user can select the version.  There is a companion command `Get-GoVersions` which will display the version numbers installed on the system.

# Side Effects
## Container size
The resulting container is actually smaller.  I don't fully understand why, as all the software as the same, but it seems to be an artifact of using a single layer rather than one layer per install.  This seems to be a good effect
## Iterating on the container.
Prior to this change, when trying to develop the container itself, then the last successful install was stored as the last layer written.  The developer could then start the shell from the last layer, and proceed to help iterate on the container faster.
Since it's all one layer, then when there's a failure, then you're back to the beginning.  This is definitely a step back.  If it's considered a serious step back, I have some ideas how to mitigate.


iglendd (approved):
I love the script and refactoring you have put in to clean it up. My concern only individual ps1 files which moved and cannot be compared (some moves are comparable for some reasons).

Then we had internal debates about pros and cons of this approach and its usage in dev cycle.

iglendd:
""My concern only individual ps1 files which moved and cannot be compared (some moves are comparable for some reasons).""
"In git the files need to be moved & committed before being modified so that it's possible to track them as being identical."
@derekwbrown if you reapply your change by explicitly moving file, may they will lineup better for the code review

derekwbrown:
> 



> My main concern with this PR is that now updating the buildimage will take much longer as any change will require to fully rebuild the container. This will also impact our CI which won't be able to reuse the cached unchanged layers. I think it's a significant downside just so that the local development environment match the build environment by running the same script. IMO there rarely is a need to have a local environment match exactly the build environment for local development and when there is we have the buildimage for that. What we can do is leave the buildimage like it is and have a script that reproduces the Dockerfile locally on a development machine, which is what @iglendd did in the first place.

So, it's not clear to me when/if we get the caching benefits, although it's true this new method reconstructs the entire container from scratch.  For comparison, this build took 1 hr; comparing to the build where we added `codeql`, which is the most charitable comparison (because it added a layer at the end so presumably got the most benefit of caching), that took 44 minutes.  So it's longer, but not considerably so (and again, that's the best case).

As for what @iglendd did, the reason that this is "better" is that what he did can only be done once. 

Again, as an example... you run what Len wrote... and then Bryce adds `ninja` for the system-probe builds.  You can't just rerun the script and get the updates.  With this you can.  Same thing when we update go versions.

I disagree with this
`IMO there rarely is a need to have a local environment match exactly the build environment for local development `
As soon as someone adds a tool, you need to match the build environment.

Ultimately, I'm (no longer) the owner of this... I put this forth as a suggestion/possibility.  If the team doesn't like it then we can just close the PR w/o merging.

KSerrania:
> This will also impact our CI which won't be able to reuse the cached unchanged layers.

The CI doesn't use the docker cache (`docker build` is run with the `--no-cache` option), so this doesn't actually change things much there (though it does impact people that are building the build image locally).

IMO, using the docker cache would have limited benefits (the buildimages jobs are run very infrequently, over a large number of runners, and the runners themselves are recycled every two days, so there's very little chance to get a cache hit), and will become a noop if/once we implement Windows runner autoscaling (since a given runner will never get reused), except if we build some kind of solution to share & manage a docker layer cache between all runners.

iglendd (approved):

windows/Dockerfile
```diff
@@ -78,147 +53,23 @@ LABEL embedded_py2_version=${EMBEDDED_PYTHON_2_VERSION}
...
-RUN [Environment]::SetEnvironmentVariable(\"Path\", [Environment]::GetEnvironmentVariable(\"Path\", [EnvironmentVariableTarget]::Machine) + \";${env:ProgramFiles(x86)}\Windows Kits\8.1\bin\x64\", [System.EnvironmentVariableTarget]::Machine)
-
-# Set 32-bit flag env var
-RUN if ($Env:TARGET_ARCH -eq 'x86') { setx WINDOWS_BUILD_32_BIT 1 }
-
+COPY ./python-packages-versions.txt /python-packages-versions.txt
 COPY ./windows/set_cpython_compiler.cmd set_cpython_compiler.cmd
-RUN .\set_cpython_compiler.cmd
+
+RUN .\install-all.ps1 -TargetContainer
```
julien-lebot, line 68: Reading this [docker documentation page](https://docs.docker.com/engine/reference/commandline/build/#squash-an-images-layers---squash-experimental) it appears that having a single layer may actually negatively impact our runners performance:

> [a single layer] image cannot take advantage of layer sharing with other images, and may use significantly more space. Sharing the base image is still supported.
While [it] may produce smaller images, it may have a negative impact on performance, as a single layer takes longer to extract, and downloading a single layer cannot be parallelized.

clarkb7: I believe this is specifically referring to squashing the resulting image into a single layer.

julien-lebot: This information applies to any image.
While having too many layers results on wasted space, too little layers impacts download and extraction parallelization and layer sharing.
It's a tradeoff between size and speed.

clarkb7: >  too little layers impacts download and extraction parallelization and layer sharing.

👍

windows/Dockerfile
```diff
@@ -78,147 +53,30 @@ LABEL embedded_py2_version=${EMBEDDED_PYTHON_2_VERSION}
...
-RUN powershell -Command .\install_winget.ps1 -Version $ENV:WINGET_VERSION -Sha256 $ENV:WINGET_SHA256
 
-# Install golang and set GOPATH to the dev path used in builds & tests
-# RUN cinst -y --no-progress golang $ENV:CHOCO_ARCH_FLAG --version $ENV:GO_VERSION
+## WORKDIR should automatically create the directory
+WORKDIR c:/scripts
+COPY ./windows/versions.ps1 versions.ps1
+COPY ./windows/install-all.ps1 install-all.ps1
+COPY ./windows/helpers.ps1 helpers.ps1
+COPY ./windows/helpers/*.ps1 c:/scripts/helpers/
```
julien-lebot, line 62:
```suggestion
COPY ./windows/* .
COPY ./windows/helpers/*.ps1 c:/scripts/helpers/
```

clarkb7:
rebasing to resolve conflicts

julien-lebot (approved):

clarkb7 (approved):

bkabrda (approved):

windows/helpers.ps1
```diff
@@ -0,0 +1,192 @@
...
+function Get-RemoteFile() {
+    param(
+        [Parameter(Mandatory = $true)][string] $RemoteFile,
+        [Parameter(Mandatory = $true)][string] $LocalFile,
+        [Parameter(Mandatory = $false)][string] $VerifyHash
+    )
+    Write-Host -ForegroundColor Green "Downloading: $RemoteFile"
+    Write-Host -ForegroundColor Green "         To: $LocalFile"
+    (New-Object System.Net.WebClient).DownloadFile($RemoteFile, $LocalFile)
+    if ($VerifyHash){
```
clarkb7, line 65:
```suggestion
    if ($PSBoundParameters.ContainsKey("VerifyHash")){
```

windows/versions.ps1
```diff
@@ -0,0 +1,44 @@
...
+    "WINGIT_SHA256"="b0442f1b8ea40b6f94ef9a611121d2c204f6aa7f29c54315d2ce59876c3d134e";
+    "SEVENZIP_VERSION"="19.0.0";
+    "SEVENZIP_SHA256"="0f5d4dbbe5e55b7aa31b91e5925ed901fdf46a367491d81381846f05ad54c45e";
+    "VS2017BUILDTOOLS_VERSION"="16.8.3.0";
+    "VS2017BUILDTOOLS_SHA256"="ccfb9355f4f753315455542f966025f96de734292d3908c8c3717e9685b709f0";
+    "VS2017BUILDTOOLS_DOWNLOAD_URL"="https://download.visualstudio.microsoft.com/download/pr/9b3476ff-6d0a-4ff8-956d-270147f21cd4/ccfb9355f4f753315455542f966025f96de734292d3908c8c3717e9685b709f0/vs_BuildTools.exe";
+    "VS2019INSTALLER_DOWNLOAD_URL"="https://download.visualstudio.microsoft.com/download/pr/3a7354bc-d2e4-430f-92d0-9abd031b5ee5/d9fc228ea71a98adc7bc5f5d8e8800684c647e955601ed721fcb29f74ace7536/vs_Community.exe";
+    "VS2019INSTALLER_SHA256"="d9fc228ea71a98adc7bc5f5d8e8800684c647e955601ed721fcb29f74ace7536";
+    "GO_VERSION"="1.19.5";
+    "GO_SHA25"="6167db91a2e40aeb453d3e59d213ecab06f62e1c4a84d13a06ccda1d999961caa";
```
clarkb7, line 14:
```suggestion
    "GO_SHA256"="167db91a2e40aeb453d3e59d213ecab06f62e1c4a84d13a06ccda1d999961caa";
```

clarkb7 (changes requested):

clarkb7 (approved):

PR #278, opened 2022-09-07
Bump docker-py version to support Cgroupns parameter

vboulineau:


mx-psi:
I think for this to work the Python version on the buildimages needs to be bumped to something more recent

vboulineau:
> I think for this to work the Python version on the buildimages needs to be bumped to something more recent

Yes, I'm on it, it requires 3.7

setup_python.sh
```diff
@@ -30,33 +30,29 @@ case $DD_TARGET_ARCH in
     ;;
 "armhf")
     detect_distro
+    echo "Installing Python from source (deb_armhf)"
```
mx-psi, line 33: Should this be?
```suggestion
    echo "Installing Python from source (armhf)"
```

mx-psi (approved):
LGTM otherwise

PR #292, opened 2022-10-01
Remove system-probe images entrypoint

paulcacheux:
system-probe images use the entrypoint to setup the gimme go version. With a potential future move to k8s runners, we will need to not use entrypoints since those are not supported. This PR removes the system probe entrypoints by installing golang manually, and setting up the `*PATH` manually as well.

Agent pipeline: https://gitlab.ddbuild.io/DataDog/datadog-agent/-/pipelines/12800264
--> the jobs to look for are `build_system_probe*`, `build_clang*`, `tests_ebpf*`, `security_go_generate_check` and dependents

### DIFF
--- a/system-probe_arm64/Dockerfile
+++ b/system-probe_arm64/Dockerfile
@@ -50,9 +50,10 @@ RUN apt-get update && apt-get install -y --no-install-recommends apt-utils && ap
 ENV GO_VERSION $GIMME_GO_VERSION
 ENV GOPATH=/go
 
-RUN wget -O /tmp/golang.tar.gz https://go.dev/dl/go$GO_VERSION.linux-arm64.tar.gz
-RUN echo "838ffa94158125f16e4aa667ee4f6b499ea57e3e35a7e2517ad357ea06714691 /tmp/golang.tar.gz" | sha256sum --check
-RUN tar -C /usr/local -xzf /tmp/golang.tar.gz
+RUN wget -O /tmp/golang.tar.gz https://go.dev/dl/go$GO_VERSION.linux-arm64.tar.gz \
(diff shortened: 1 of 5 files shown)


system-probe_arm64/Dockerfile
```diff
@@ -50,16 +47,16 @@ RUN apt-get update && apt-get install -y --no-install-recommends apt-utils && ap
         wget \
         xz-utils
 
-RUN wget -O /bin/gimme https://raw.githubusercontent.com/travis-ci/gimme/v1.5.4/gimme
-RUN echo "03b295636d4e22870b6f6e9bc06a71d65311ae90d3d48cbc7071f82dd5837fbc  /bin/gimme" | sha256sum --check
-RUN chmod +x /bin/gimme
-RUN gimme $GIMME_GO_VERSION
+ENV GO_VERSION $GIMME_GO_VERSION
```
brycekahle, line 50: Is `GO_VERSION` needed for someone else, or could you use `GIMME_GO_VERSION` in the download command below?

paulcacheux: And I wanted to keep the ENV=ARG pattern that was here before as well

brycekahle: Should we replace `ARG GIMME_GO_VERSION` with `ARG GO_VERSION` then? You cannot really use a different version from the command line because the hashes won't match.

paulcacheux: Yeah that's a good point, let me do that

brycekahle (approved):

paulcacheux:
Thanks, I will merge this once the freeze on the agent side is lifted

bkabrda (approved):

PR #305, opened 2022-10-18
Create btf-gen image

ISauve:
Adds a build image to be used for generating minimized BTFs for our bpf programs in our CI.
This job needs its own image because it requires a new version of `bpftool` which isn't available on the distributions our other build images are running.


btf-gen/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM debian:bookworm
```
KSerrania, line 1: Would you mind setting your team as the code owner of this file?

KSerrania (approved):

PR #306, opened 2022-10-19
Revert "Bump kitchen dependencies"

KSerrania:
Reverts DataDog/datadog-agent-buildimages#259 while we investigate why kitchen tests in datadog-agent are failing.


KSerrania:
This will probably fix the issue without having to revert: https://github.com/DataDog/datadog-agent/compare/kserrania/fix-kitchen

PR #307, opened 2022-10-25
Update MSYS to 20230318

julien-lebot:
Update MSYS to version `20230318`.
For some reason `ridk install 3` includes several `automake` versions, including some that don't seem to exists. I replaced the command with its equivalent, but without the failing automake versions.


iglendd (approved):

chouetz (approved):
ok

PR #323, opened 2022-11-30
Move deb-x64 builder to Ubuntu 14.04

bkabrda:
Ubuntu 14.04 still has extended LTS support, while Debian 8 hasn't been supported for quite a while and its repodata signing keys have expired as a result of that.


truthbk:
@bkabrda after merging the latest changes where we build patchelf from source our build issues were solved for the datadog-fips-proxy project. So, feel free to merge this whenever you're ready. Some conflcits, but otherwise we're looking set.

truthbk (approved):

KSerrania (approved):

PR #332, opened 2022-12-22
Allow building openscap using omnibus

lebauce:
This PR makes it possible to compile openscap using omnibus.

In particular, it:
- compiles and installs gcc 8.4.0 to be able to compile
APT and openscap which require a modern C++ compiler.
- installs triehash - to compile APT - which is only available as a package
on Ubuntu and not available any more on latest LTS.
- installs flex to compile RPM

### DIFF
--- a/build-gcc.sh
+++ b/build-gcc.sh
@@ -1,0 +1,56 @@
+#!/bin/bash
+
+set -e
+
+GCC_VERSION=8.4.0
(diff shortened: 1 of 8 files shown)


jinroh:
rebased on 801f6e2e9d5ffb350887dc8ff49aab72ea7d8d24

PR #335, opened 2023-01-01
[go] bumping to 1.19.4

truthbk:
Bumps Go to version 1.19.4.

Notes: the go `1.19.4` build fails on `deb-x64` with the old `ld` linker, and thus an updated binutils is installed from source to workaround the issue. It's a well-known issue with go`-race` builds and older linkers.


paulcacheux (approved):

brycekahle (approved):

setup_go.sh
```diff
@@ -7,20 +7,50 @@ set -ex
...
+
 # Install gimme to get go1.15.11
-curl -sL -o /bin/gimme https://raw.githubusercontent.com/travis-ci/gimme/v1.5.4/gimme
+curl -sL -o /bin/gimme https://raw.githubusercontent.com/travis-ci/gimme/v1.5.5/gimme
 # Check sha256sum, fail otherwise
-echo "03b295636d4e22870b6f6e9bc06a71d65311ae90d3d48cbc7071f82dd5837fbc  /bin/gimme" | sha256sum --check
+echo "3d565d57ec28edb14be9e540cd3e628607ec5b791e78224c47250d36ce4aedf2  /bin/gimme" | sha256sum --check
 chmod +x /bin/gimme
-eval "$(gimme 1.15.11)"
+eval "$(gimme 1.18.9)"
```
bkabrda, line 33: Why 1.18.9 and not 1.19.4?

truthbk: It's the bootstrap golang; I am OK with setting it to `1.19.4`, I juste operated under the assumption here that there was a reason why we wanted to bootstrap with a prior version.

rpm-x64/Dockerfile
```diff
@@ -44,11 +49,19 @@ ENV RUSTC_SHA256 $RUSTC_SHA256
...
 ENV PERL_VERSION $PERL_VERSION
+ENV BINUTILS_VERSION $BINUTILS_VERSION
+ENV BINUTILS_SHA256 $BINUTILS_SHA256
+ENV BASE_IMAGE $BASE_IMAGE
+
 
 # Enable the vault (archive) repos, as we are past CentOS6 EOL
-RUN sed -is 's/enabled=0/enabled=1/g' /etc/yum.repos.d/CentOS-Vault.repo && \
+RUN if [[ $(echo $BASE_IMAGE | cut -f2 -d: | head -c1) == 6 ]]; then \ 
+      echo "VICTORY"; fi
```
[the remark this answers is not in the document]
truthbk: lol, yeah.... I was troubleshooting the BASE_IMAGE build arg. Thanks for spotting this; I actually thought I had removed it already.

setup_go.sh
```diff
@@ -7,20 +7,50 @@ set -ex
...
 # Update PATH to include the built go binaries
 echo 'export PATH="/goroot/bin:$PATH"' >> /root/.bashrc
 
+# Remove bison sources
+rm -rf bison-${BISON_VERSION}
+rm -rf bison-${BISON_VERSION}.tar.gz
+
+# Remove binutils sources
+rm -rf binutils-${BINUTILS_VERSION}
+rm -rf binutils-${BINUTILS_VERSION}.tar.gz
```
sgnn7, line 52: nit: All of the `rm -rf`s probably need double-quotes around the args for safety
nit: these probably would be fine to co-locate w/ install parts so that maintenance is easier

.gitlab-ci.yml
```diff
@@ -201,7 +209,7 @@ trigger_build_kernels:
   extends: .test
   before_script:
     - rm -r /go/src/github.com/DataDog/datadog-agent || true
-    - git clone https://github.com/DataDog/datadog-agent /go/src/github.com/DataDog/datadog-agent
+    - git clone -b $AGENT_TESTING_BRANCH https://github.com/DataDog/datadog-agent /go/src/github.com/DataDog/datadog-agent
```
sgnn7, line 212: I'm guessing this will be removed after the corresponding agent PR is merged?

deb-x64/Dockerfile
```diff
@@ -24,6 +24,10 @@ ARG RUSTUP_VERSION=1.24.3
 ARG RUSTUP_SHA256="3dc5ef50861ee18657f9db2eeb7392f9c2a6c95c90ab41e45ab4ca71476b4338"
 ARG PATCHELF_VERSION=0.13.1
 ARG PATCHELF_SHA256="f6d5ecdb51ad78e963233cfde15020f9eebc9d9c7c747aaed54ce39c284ad019"
+ARG BISON_VERSION="3.8"
+ARG BISON_SHA256="d5d184d421aee15603939973a6b0f372f908edfb24c5bc740697497021ad9458"
+ARG BINUTILS_VERSION="2.39"
+ARG BINUTILS_SHA256="d12ea6f239f1ffe3533ea11ad6e224ffcb89eb5d01bbea589e9158780fa11f10"
```
sgnn7, line 30: nit: Not a blocker but this block of `ARG`s is really hard to visually parse quickly - we may want to maybe use the hanging defs to make it easier to read:
```dockerfile
ARG FOO="3.2.1" \
    FOO_SHA="abc123"
ARG BAR="1.2.3" \
    BAR_SHA="abc123"
```

rpm-x64/Dockerfile
```diff
@@ -44,11 +49,18 @@ ENV RUSTC_SHA256 $RUSTC_SHA256
...
+ENV BINUTILS_VERSION $BINUTILS_VERSION
+ENV BINUTILS_SHA256 $BINUTILS_SHA256
+ENV BASE_IMAGE $BASE_IMAGE
+
+# persist RHEL major for readable output
+RUN echo $(cat /etc/redhat-release | cut -d'.' -f1 | awk '{print $NF}') > /etc/redhat-release-major
 
 # Enable the vault (archive) repos, as we are past CentOS6 EOL
-RUN sed -is 's/enabled=0/enabled=1/g' /etc/yum.repos.d/CentOS-Vault.repo && \
+RUN if [[ $(cat /etc/redhat-release-major) == 6 ]]; then \
```
sgnn7, line 60: This check is reused 5+ times in the file. Can we set a variable/func and then check that instead?
```suggestion
RUN if [[ "$(cat /etc/redhat-release-major)" == 6 ]]; then \ 
```

sgnn7: nit: Do we want to print out some debug info if we end up in this block? If so - we may want to do that for other similar blocks.

setup_go.sh
```diff
@@ -7,20 +7,50 @@ set -ex
...
+echo "${BISON_SHA256}  ./bison-${BISON_VERSION}.tar.gz" | sha256sum --check
+tar -zxvf ./bison-${BISON_VERSION}.tar.gz
+cd bison-${BISON_VERSION}
+./configure --prefix=/usr/local/bison && make && make install
+cd -
+
+export PATH=/usr/local/bison/bin:$PATH
+
+# Upgrade binutils
+curl -sL -O https://ftp.gnu.org/gnu/binutils/binutils-${BINUTILS_VERSION}.tar.gz
```
sgnn7, line 21: nit. ditto on other url.
```suggestion
curl -sL -O "https://ftp.gnu.org/gnu/binutils/binutils-${BINUTILS_VERSION}.tar.gz"
```

sgnn7 (approved):
LGTM - mainly left some non-blocking nit comments.

.gitlab-ci.yml
truthbk: This will be kept but the variable at the top will be set to `main`

deb-x64/Dockerfile
truthbk: This already took this form; I'd rather address additional changes to the dockerfile in a separate PR that might address maintainability issues.

setup_go.sh
bkabrda: You're right, feel free to disregard the comment.

bkabrda (approved):

PR #345, opened 2023-01-25
Add dependencies for kernel version testing stage

usamasaqib:


ISauve (approved):

system-probe_arm64/Dockerfile
```diff
@@ -48,7 +48,9 @@ RUN apt-get update && apt-get install -y --no-install-recommends apt-utils && ap
         python3-dev \
         default-jre \
         wget \
-        xz-utils
+        xz-utils \
+        openssh-client \
```
brycekahle, line 52: nit: I think the package list was intended to be alphabetical. That way it is easier to see if a package is already included.

system-probe_x64/Dockerfile
```diff
@@ -72,6 +74,14 @@ COPY ./requirements.txt /
 RUN python3 -m pip install wheel
 RUN python3 -m pip install -r requirements.txt
 
+# Install pulumi
+RUN curl -fsSL https://get.pulumi.com > install-pulumi.sh
+RUN echo "060d39770476f7d11de7c75dc1e1780f6492455555b36f4d3ff18f89752ebfc6  install-pulumi.sh" | sha256sum --check
```
KSerrania, line 79: This will break when the install script at that URL changes. Is there perhaps a versioned install script available somewhere?

That's what we do with the Rust install script: instead of fetching https://sh.rustup.rs, we use a specific version: https://github.com/DataDog/datadog-agent-buildimages/blob/c5447ff4fc2d639923198fc954fd43d2cb7e5d88/deb-x64/Dockerfile#L168

If there's no such thing available for Pulumi, this is not a big deal, we'll just need to fix the Dockerfile when the script gets updated.

KSerrania (approved):

system-probe_x64/Dockerfile
brycekahle: It appears to be here: https://github.com/pulumi/get.pulumi.com/blob/master/dist/install.sh

brycekahle: We should probably specify the pulumi version with `--version`, otherwise it will always download the latest.

Alternatively, they have tarballs available with manual install instructions.

PR #346, opened 2023-01-31
Break install-all into phases to allow for some Docker layers

clarkb7:
Break install-all into phases to allow for some Docker layers

Keep `main` behavior in env var helpers and write to USER env instead of SYSTEM env
* having git find on SYSTEM path broke `vs_buildtools.exe`


windows/Dockerfile
```diff
@@ -65,7 +65,14 @@ COPY ./requirements.txt ./requirements-py2.txt /
...
 
-RUN .\install-all.ps1 -TargetContainer
+COPY ./windows/helpers/phase1/*.ps1 c:/scripts/helpers/phase1/
+RUN .\install-all.ps1 -TargetContainer -Phase 1
+COPY ./windows/helpers/phase2/*.ps1 c:/scripts/helpers/phase2/
+RUN .\install-all.ps1 -TargetContainer -Phase 2
+COPY ./windows/helpers/phase3/*.ps1 c:/scripts/helpers/phase3/
+RUN .\install-all.ps1 -TargetContainer -Phase 3
+COPY ./windows/helpers/phase4/*.ps1 c:/scripts/helpers/phase4/
+RUN .\install-all.ps1 -TargetContainer -Phase 4
```
iglendd, line 75: Is it possible to have a more meaningful names instead of numbers (I wonder)

iglendd (approved):

julien-lebot (approved):
👍

PR #351, opened 2023-02-08
Install GCC 10.4.0

0intro:
This change installs GCC 10.4.0 on the following build images:

 - deb-arm
 - deb-x64
 - rpm-arm64
 - rpm-x64
 - suse-x64

We chose to provide our own GCC version, because the GCC versions available as part of CentOS 6, openSUSE 42.1 and Ubuntu 14.04 weren't recent enough to build OpenSCAP and some of its dependencies.

We chose GCC 10.4.0, because it was the most recent version that could be built using the GCC versions available on our distributions. GCC 11 and upper include libcody, which is written in C++11 and requires a more recent GCC version to be built.

This change is the result of a collaborative work between Sylvain Baubeau, Pierre Guilleminot and David du Colombier.

### DIFF
--- a/build-gcc.sh
+++ b/build-gcc.sh
@@ -1,0 +1,41 @@
+#!/bin/bash
+
+set -e
+
+if [ -z "$GCC_VERSION" ] || [ -z "$GCC_SHA256" ]; then
+	echo "The GCC_VERSION and GCC_SHA256 environment variables should be defined to run this script" >&2
+	exit 1
+fi
+
+PREFIX=/opt/gcc-${GCC_VERSION}
+
+compile_with_autoconf() {
+	LIB=lib
+	if [[ "$DD_TARGET_ARCH" == *64* ]]; then
+		LIB=lib64
+	fi
+	[ -e /etc/redhat-release ] && configure_args="--libdir=$PREFIX/$LIB" || true
+	./configure --prefix=$PREFIX $configure_args $*
+	cpu_count=$(grep process /proc/cpuinfo | wc -l)
+	make -j $cpu_count --silent
+	make install
+}
+
+url="https://mirrors.kernel.org/gnu/gcc/gcc-${GCC_VERSION}/gcc-${GCC_VERSION}.tar.gz"
+archive=$(basename $url)
+[ ! -e "$archive" ] && curl -LO $url || true
+echo "${GCC_SHA256}  ${archive}" | sha256sum --check
+
+tar xzf $(basename $url)
+cd "gcc-${GCC_VERSION}"
+
+contrib/download_prerequisites
+
+compile_with_autoconf \
+    --disable-nls \
+    --enable-languages=c,c++ \
+    --disable-multilib
+
+cd -
+
+rm -rf "gcc-${GCC_VERSION}" "gcc-${GCC_VERSION}.tar.gz"
--- a/deb-arm/Dockerfile
+++ b/deb-arm/Dockerfile
@@ -22,7 +22,8 @@ ARG CMAKE_SHA256="9f8d42ef0b33d1bea47afe15875435dac58503d6a3b58842b473fd811e6df1
 ARG CLANG_VERSION=8.0.0
 ARG CLANG_SHA256="a77eb8fde0a475c25d46dccdeb851a83cbeeeb11779fa2218ae19db9cd0e51f9"
 ARG DD_TARGET_ARCH=aarch64
-
+ARG GCC_VERSION=10.4.0
+ARG GCC_SHA256=ab1974017834430de27fd803ade4389602a7d6ca1362496c57bef384b2a4cb07
 
 # Environment
 ENV GOPATH /go
@@ -33,6 +34,7 @@ ENV CLANG_VERSION $CLANG_VERSION
 ENV CLANG_SHA256 $CLANG_SHA256
 ENV CONDA_PATH /root/miniforge3
 ENV DD_TARGET_ARCH $DD_TARGET_ARCH
+ENV GCC_VERSION $GCC_VERSION
 
 # Remove the early return on non-interactive shells, which makes sourcing the file not activate conda
 RUN grep -v return /root/.bashrc >> /root/newbashrc && cp /root/newbashrc /root/.bashrc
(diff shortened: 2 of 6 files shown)


bkabrda:
:wave: thanks for submitting the PR. Have you talked about this to other teams that maintain pieces of Agent code that get compiled by GCC? From top of my head, this would be agent-metrics-logs (rtloader) and agent-integrations (some Python integrations have C extensions that would get compiled by GCC). It would be good to try to standardize on one GCC version instead of adding more, if possible.

I'm also wondering whether it's practical to put the new GCC version under `/opt/gcc-${GCC_VERSION}`, because anything using this GCC will have to hardcode the specific `GCC_VERSION` and if we do even a minor version upgrade we'll have to fix all these places. Maybe it would be sufficient to have a path that would only contain GCC minor version, e.g. `/opt/gcc-10`?

paulcacheux (approved):

safchain (approved):

0intro:
Thanks for your review.

> wave thanks for submitting the PR. Have you talked about this to other teams that maintain pieces of Agent code that get compiled by GCC? From top of my head, this would be agent-metrics-logs (rtloader) and agent-integrations (some Python integrations have C extensions that would get compiled by GCC). It would be good to try to standardize on one GCC version instead of adding more, if possible.

We have not talked to the other teams yet. Our goal was to introduce a recent GCC version in the build image, which is required to build some external dependencies (for example, libraries written in C++11). Then, the other teams will be able to use this recent GCC version.

Currently, the build of the Agent is relying on heterogeneous GCC toolchains provided as part of the various Linux distribution we support (RHEL, Ubuntu and SUSE). I think the ultimate goal is that everyone use a single, common, recent GCC toolchain we build ourself, instead of relying on the heterogeneity of the various Linux distribution.

> I'm also wondering whether it's practical to put the new GCC version under `/opt/gcc-${GCC_VERSION}`, because anything using this GCC will have to hardcode the specific `GCC_VERSION` and if we do even a minor version upgrade we'll have to fix all these places. Maybe it would be sufficient to have a path that would only contain GCC minor version, e.g. `/opt/gcc-10`?

The `GCC_VERSION` variable is exported to the environment of the build image, so, when building the Agent, one could refer to the `GCC_VERSION` available.

It would be possible, for example, to export something like `GCC_10_VERSION`, or, as you suggested, use a fixed path `/opt/gcc-10`, but I think it would defeat the goal of having a single, common, modern, recent GCC version.
We defined a `GCC_VERSION` variable, like it was the case, for example, for `CMAKE_VERSION` already, but I think, similarly to what you suggested, that we could simply hardcode the path to something like `/opt/gcc`.

What do you think?

build-gcc.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
+
+set -e
+
+GCC_VERSION=10.4.0
+GCC_SHA256=ab1974017834430de27fd803ade4389602a7d6ca1362496c57bef384b2a4cb07
+PREFIX="${PREFIX:-/usr}"
```
bkabrda, line 7: AFAICS all Dockerfiles set this to `/opt/gcc-${GCC_VERSION}`. Would it make sense to make that value default and not require passing it explicitly?

build-gcc.sh
```diff
@@ -0,0 +1,58 @@
...
+        echo "Unsupported archive format $extension"
+        exit 1
+        ;;
+    esac
+
+    tar x${tar_arg}f ${archive}
+}
+
+compile_with_autoconf() {
+    [ -e /etc/redhat-release ] && configure_args="--libdir=$PREFIX/lib64" || true
```
bkabrda, line 34: I think there should be additional check here to ensure the architecture is 64bit. We also have `rpm-armhf` (32bit ARM) builder where this shouldn't apply.

Is there a specific reason why this feature is not added to the `rpm-armhf` builder?

bkabrda (changes requested):
Hi :wave: the PR mostly LGTM, but I left couple notes and comments inline to get addressed.

build-gcc.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
+
+set -e
+
+GCC_VERSION=10.4.0
```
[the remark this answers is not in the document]
0intro: I've removed `GCC_VERSION` and `GCC_SHA256` and added a check to ensure these variables have been defined.

build-gcc.sh
0intro: I've set `PREFIX=/opt/gcc-${GCC_VERSION}` in `build-gcc.sh`.

build-gcc.sh
```diff
@@ -0,0 +1,58 @@
...
+GCC_SHA256=ab1974017834430de27fd803ade4389602a7d6ca1362496c57bef384b2a4cb07
+PREFIX="${PREFIX:-/usr}"
+
+prepare_sources() {
```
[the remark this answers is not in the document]
0intro: I've removed the `prepare_sources` function and now call `tar` directly.

build-gcc.sh
0intro: I've added a check to set up `--libdir=$PREFIX/lib64` or `--libdir=$PREFIX/lib` according to the architecture.

We have not added GCC in the `rpm-armhf` builder yet because we experienced a build issue:

```
checking for aarch64-unknown-linux-gnu-gcc... /gcc-10.4.0/host-aarch64-unknown-linux-gnu/gcc/xgcc -B/gcc-10.4.0/host-aarch64-unknown-linux-gnu/gcc/ -B/opt/gcc-10.4.0/aarch64-unknown-linux-gnu/bin/ -B/opt/gcc-10.4.0/aarch64-unknown-linux-gnu/lib/ -isystem /opt/gcc-10.4.0/aarch64-unknown-linux-gnu/include -isystem /opt/gcc-10.4.0/aarch64-unknown-linux-gnu/sys-include   -fno-checking
checking for suffix of object files... configure: error: in `/gcc-10.4.0/aarch64-unknown-linux-gnu/libgcc':
configure: error: cannot compute suffix of object files: cannot compile

bkabrda (approved):
The change LGTM, so I'm approving.

The two points that I'll leave up to you folks to address going forward are:
* The rpm-armhf addition of this gcc version when you figure out how to fix the compilation problem there.
* Talking to other Agent teams to consolidate on this new gcc version and stop using the old ones.

jinroh:
rebased on 48c12a15b9c586bc22fd248a74a71cc355bd9af5

0intro:
I split the addition of flex on the images where it was missing to a different commit.

0intro:
Moved the addition of flex to a separate PR (#371).

PR #356, opened 2023-02-14
[CI] Use `_test_only` ECR for commits not made on `main`

amenasria:
This PR stores the docker images in different ECR based on where the commit was made:
- Docker images made from a commit on `main` will be stored in `ci/datadog-agent-buildimages/$IMAGE`
- Docker images made from a commit not on `main` will be stored in `datadog-agent-buildimages/$IMAGE_test_only`

We're doing this because the increased activity on this repo made the former ECR reach its image number limit quicker, making some images expire before we stopped using them in the CI.


bkabrda (changes requested):
You also need to adapt all the `test_*` jobs to test the images with the `_test_only` suffix. I also think that you need to do the same set of changes to windows build+test jobs.

.gitlab-ci.yml
```diff
@@ -45,22 +46,27 @@ default:
 
 .build:
   stage: build
-  except: [ tags ]
+  rules:
+    - if: '$CI_COMMIT_TAG == null'
+    # If the commit was made on main then use the classic ECR
+    - if: '$CI_COMMIT_BRANCH == "main"'
+      variables:
+        ECR_TEST_ONLY: ""
```
KSerrania, line 54: Rules are [evaluated in order](https://docs.gitlab.com/ee/ci/yaml/#rules), and only the first matching rule is used. Since `$CI_COMMIT_BRANCH == "main"` => `$CI_COMMIT_TAG == null`, the second case will never get matched and `ECR_TEST_ONLY` will always be `_test_only`.

.gitlab-ci.yml
```diff
@@ -219,37 +225,37 @@ trigger_build_kernels:
 
 test_deb_x64:
   extends: .test_agent6
-  image: 486234852809.dkr.ecr.us-east-1.amazonaws.com/ci/datadog-agent-buildimages/deb_x64:v$CI_PIPELINE_ID
+  image: 486234852809.dkr.ecr.us-east-1.amazonaws.com/ci/datadog-agent-buildimages/deb_x64${ECR_TEST_ONLY}:v$CI_PIPELINE_ID
```
KSerrania, line 228: Tests are going to fail on `main` with that setup, as `ECR_TEST_ONLY` is not set to the empty string on `main` pipelines for this job (and the other test jobs).

KSerrania (commented):
How about defining `ECR_TEST_ONLY` at the workflow level? See this example in datadog-agent: https://github.com/DataDog/datadog-agent/blob/db8e9b9079c334c90278a4d95d74f83b55779b11/.gitlab-ci.yml#L265-L278

You could define  the `TEST_ECR_ONLY` variable once there (with different values depending on the matched rule), and that would be set for all jobs in the pipeline.

bkabrda (approved):

amenasria:
Just added the permissions for windows runners to be able to push the docker images to the ECR !

PR #370, opened 2023-03-05
Remove entrypoint from all Agent build images

KSerrania:
### What does this PR do?

Updates the deb-x64, deb-arm64, deb-armhf, rpm-x64, rpm-arm64, rpm-armhf, suse-x64 images and removes their entrypoint.

### Motivation

Prepare the migration to kubernetes-based Gitlab runners, which overwrite the entrypoint of images used in jobs.

### Additional notes / Follow-ups

For now, jobs using these images need to run `source /root/.bashrc` as their first step, to run the setup `rvm` scripts and use the `ddpy3` conda environment.

Therefore, these new images require changes on the `datadog-agent` side: https://github.com/DataDog/datadog-agent/compare/kserrania/buildimages-remove-entrypoint.

The needed `source /root/.bashrc` call could be removed later, if we manage to:
- use `base` as the Python 3 environment instead of a separate `ddpy3` environment (which would remove the need to do `conda activate ddpy3` in `/root/.bashrc`),
- install `ruby` without `rvm` (which would remove the need to do `source /usr/local/rvm/scripts/rvm` in `/root/.bashrc`),
- find a way to direct `CC` / `CPP` / `CXX` to the devtoolset without using `/root/.bashrc` [in the rpm-x64 image](https://github.com/DataDog/datadog-agent-buildimages/blob/891e1344c64d7c44e58b5d08c04ebed970552dd6/rpm-x64/Dockerfile#L229-L234).

Test Agent pipeline: https://gitlab.ddbuild.io/DataDog/datadog-agent/pipelines/13858876


paulcacheux:
Regarding the rvm install I think we could build ruby from source, using instructions from https://github.com/ruby/ruby/blob/master/doc/contributing/building_ruby.md. This is basically what rvm is doing for us. I can do a PR based on this branch with my suggested changes

KSerrania:
> Regarding the rvm install I think we could build ruby from source, using instructions from https://github.com/ruby/ruby/blob/master/doc/contributing/building_ruby.md. This is basically what rvm is doing for us. I can do a PR based on this branch with my suggested changes

Yes, that's what I was thinking about (but I haven't investigated yet). If you already have changes in flight for that, feel free to propose them and we'll add them here 👍

paulcacheux:
Regarding the gcc, CC, CPP, CXX issue do you think we could use the new gcc that we now have with https://github.com/DataDog/datadog-agent-buildimages/pull/351 ? This would also mean using the exact same gcc everywhere

deb-arm/Dockerfile
chouetz, line 1: Any reason why rpm is split between arm64 and armhf, and not the same for deb?

chouetz (approved):
OK for me

PR #372, opened 2023-03-09
remove rvm and build ruby from source

paulcacheux:
This PR removes the use of rvm to install ruby and simply build ruby from source instead. This means that ruby is directly in path and removes one of the dependency on the entrypoint scripts

This PR does not migrate the omnibus images, because they are not as important (only used by a manual job), and because we are in the process of removing them

Agent pipeline: https://gitlab.ddbuild.io/DataDog/datadog-agent/-/pipelines/13956713

### DIFF
--- a/deb-arm/Dockerfile
+++ b/deb-arm/Dockerfile
@@ -63,6 +63,7 @@ COPY ./requirements.txt ./requirements-py2.txt /
 RUN ./setup_python.sh
 ENV PATH "${CONDA_PATH}/condabin:${PATH}"
 ENV PKG_CONFIG_LIBDIR "${PKG_CONFIG_LIBDIR}:${CONDA_PATH}/lib/pkgconfig"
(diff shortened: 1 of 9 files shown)


deb-arm/Dockerfile
```diff
@@ -42,7 +42,7 @@ RUN grep -v return /root/.bashrc >> /root/newbashrc && cp /root/newbashrc /root/
 RUN apt-get update && apt-get install -y fakeroot curl git procps bzip2 \
     build-essential pkg-config tar libsystemd-dev libkrb5-dev \
     gettext libtool autopoint autoconf libtool-bin \
-    selinux-basics default-jre flex
+    selinux-basics default-jre flex libssl-dev
```
paulcacheux, line 45: This is required to have it work on arm64. It was actually installed before by the `rvm requirements` step, and is installed in the `deb-x64` images

PR #375, opened 2023-03-14
bump rpm version to 4.14.3 on openSUSE images allowing sha256 signing

Pythyu:
Changes : 
  - openSUSE rpm version bump to 4.14.3.
  - allows openSUSE to sign rpms in sha256

### DIFF
--- a/suse-x64/Dockerfile
+++ b/suse-x64/Dockerfile
@@ -161,7 +161,6 @@ RUN chmod +x /build-gcc.sh \
 RUN curl -sL -o /tmp/rpm-4.14.3.tar.bz2 http://ftp.rpm.org/releases/rpm-4.14.x/rpm-4.14.3.tar.bz2 \
     && cd /tmp \
     && echo "13711268181a6e201eb9d4eb06384a7fcc42dcc6c9057a62c53472cfc5ba44b5  rpm-4.14.3.tar.bz2" | sha256sum --check \
-    && cp /usr/include/linux/magic.h /usr/include/magic.h \
     && cp /usr/include/db4/db.h /usr/include/db.h \
     && tar -xjf /tmp/rpm-4.14.3.tar.bz2 \
     && cd rpm-4.14.3 \
--- a/suse-x64/Dockerfile
+++ b/suse-x64/Dockerfile
@@ -160,8 +160,7 @@ RUN chmod +x /build-gcc.sh \
 # Actually build new rpm
 COPY patches/rpm-4.15.1-fix-rpmbuild-segfault.patch /tmp
 # Cannot use HTTPS here: cert name is invalid
-RUN if [[ "$(cat /etc/redhat-release-major)" == 6 ]]; then \ 
-    curl -sL -o /tmp/rpm-4.15.1.tar.bz2 http://ftp.rpm.org/releases/rpm-4.15.x/rpm-4.15.1.tar.bz2 \
+RUN curl -sL -o /tmp/rpm-4.15.1.tar.bz2 http://ftp.rpm.org/releases/rpm-4.15.x/rpm-4.15.1.tar.bz2 \
     && echo "ddef45f9601cd12042edfc9b6e37efcca32814e1e0f4bb8682d08144a3e2d230  /tmp/rpm-4.15.1.tar.bz2" | sha256sum --check \
     && cd /tmp \
     && tar -xjf /tmp/rpm-4.15.1.tar.bz2 \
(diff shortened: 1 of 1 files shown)


suse-x64/Dockerfile
```diff
@@ -157,6 +158,28 @@ RUN chmod +x /build-gcc.sh \
     && /build-gcc.sh \
     && rm /build-gcc.sh
 
+RUN curl -sL -o /tmp/rpm-4.14.3.tar.bz2 http://ftp.rpm.org/releases/rpm-4.14.x/rpm-4.14.3.tar.bz2 \
+    && cd /tmp \
+    && echo "13711268181a6e201eb9d4eb06384a7fcc42dcc6c9057a62c53472cfc5ba44b5  rpm-4.14.3.tar.bz2" | sha256sum --check \
+    && cp /usr/include/db4/db.h /usr/include/db.h \
```
bkabrda, line 164: Is this necessary if we're adding `-I/usr/include/db4` to CPPFLAGS below?

suse-x64/Dockerfile
```diff
@@ -157,6 +158,28 @@ RUN chmod +x /build-gcc.sh \
...
+    && tar -xjf /tmp/rpm-4.14.3.tar.bz2 \
+    && cd rpm-4.14.3 \
+    && curl -sL -o db-4.5.20.tar.gz http://download.oracle.com/berkeley-db/db-4.5.20.tar.gz \
+    && echo "f52cd5cea899823dd200d56556f70b33c55e48a33bb7b65ee128968dc10ca82d  db-4.5.20.tar.gz" | sha256sum --check \
+    && tar xzf db-4.5.20.tar.gz \
+    && ln -s db-4.5.20 db \
+    && cd db/build_unix/ \
```
bkabrda, line 171: Why are we downloading another db4 here? I don't see it installed/used anywhere

suse-x64/Dockerfile
```diff
@@ -157,6 +158,28 @@ RUN chmod +x /build-gcc.sh \
...
+    && ../../configure --with-posixmutexes  --without-lua CPPFLAGS="-I/usr/include/nspr4 -I/usr/include/nss3 -I/usr/include/db4/"\
+    && make \
+    && make install \
+    && cd / \
+    && echo "/usr/local/lib/" >> /etc/ld.so.conf \
+    && ldconfig
```
bkabrda, line 177: What is the purpose of the two above lines? AFAICS `/usr/local/lib` already is in `/etc/ld.so.conf` in the base image.

bkabrda (changes requested):
Hi :wave: the PR mostly looks good, but I left some comments/questions inline to get addressed.

suse-x64/Dockerfile
Pythyu: It's needed for RPM to build he access his db with the ``db`` symlink. To get more info you can look at the INSTALL file in the source

suse-x64/Dockerfile
Pythyu: I'm going to check to be sure

suse-x64/Dockerfile
```diff
@@ -157,6 +158,28 @@ RUN chmod +x /build-gcc.sh \
...
+
+# Rebuild RPM database with the new rpm
+RUN mkdir -p /usr/local/var/lib/rpm \
+    && cp /var/lib/rpm/Packages /usr/local/var/lib/rpm/Packages
```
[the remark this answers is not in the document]
Pythyu: Yep fixing that

suse-x64/Dockerfile
bkabrda: From what I understood in the `INSTALL` file, you *either* use an embedded BDB (for which you'd need to download) *or* use a system installation of it (`libdb4_8`, this would probably be preferable in this case).

Pythyu: Ok yea it makes sense

suse-x64/Dockerfile
bkabrda: Interesting, in the base image, I'm getting
```
# ld --verbose | grep SEARCH
SEARCH_DIR("/usr/x86_64-suse-linux/lib64"); SEARCH_DIR("/usr/lib64"); SEARCH_DIR("/usr/local/lib64"); SEARCH_DIR("/lib64"); SEARCH_DIR("/usr/x86_64-suse-linux/lib"); SEARCH_DIR("/usr/local/lib"); SEARCH_DIR("/lib"); SEARCH_DIR("/usr/lib");
```
I'm wondering if there's something that changes that during the image build. I can't see anything that would do that from brief look :thinking:

suse-x64/Dockerfile
Pythyu: and yea it was too much, removed it

suse-x64/Dockerfile
Pythyu: Uhm I don't know can't find anything that changes it 🤔 
But without it doesn't find it and with it does do I keep it this way ?

PR #378, opened 2023-04-12
Install triehash to compile APT

lebauce:


rpm-x64/Dockerfile
```diff
@@ -280,6 +280,9 @@ RUN chmod +x /build-gcc.sh \
     && /build-gcc.sh \
     && rm /build-gcc.sh
 
+# triehash
+COPY ./triehash.pl /usr/local/bin/triehash
```
bkabrda, line 284: It looks like we're missing the `chmod` call in this dockerfile.

PR #390, opened 2023-05-19
Build circleci image with other agent buildimages

chouetz:
Moved CircleCI image definition and construction alongside other buildimages used during DataDog Agent CI.
This will reduce the risk of misalignment between the various buildimages we have and might help improving efficiency of image modifications

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -207,6 +207,12 @@ build_kernel_version_testing_arm64:
     DOCKERFILE: kernel-version-testing/kernel-version-testing_arm64/Dockerfile
     IMAGE: kernel-version-testing_arm64
 
+build_circleci_runner:
+  extends: [ .build, .x64 ]
+  variables:
+    DOCKERFILE: circleci/Dockerfile
+    IMAGE: circleci-runner
+
 trigger_build_kernels:
   stage: trigger_build_kernels
   needs: [ "build_kernel_version_testing_x64" , "build_kernel_version_testing_arm64" ]
@@ -483,6 +489,12 @@ release_windows_ltsc2022_x64:
     IMAGE: windows_ltsc2022_x64
     DOCKERHUB_IMAGE: agent-buildimages-windows_x64
     DOCKERHUB_TAG_PREFIX: "ltsc2022"
+
+release_circleci_runner:
+  extends: .release
+  needs: [ "build_circleci_runner" ]
+  variables:
+    IMAGE: circleci-runner
 
 notify-on-failure:
   extends: .slack-notifier-base
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -208,10 +208,19 @@ build_kernel_version_testing_arm64:
     IMAGE: kernel-version-testing_arm64
 
 build_circleci_runner:
-  extends: [ .build, .x64 ]
+  extends: [ .x64 ]
+  stage: build
+  except: [ tags ]
   variables:
     DOCKERFILE: circleci/Dockerfile
     IMAGE: circleci-runner
+  script:
+    # Dockerhub login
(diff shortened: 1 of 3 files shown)


circleci/requirements.txt
```diff
@@ -0,0 +1,9 @@
+wheel
```
chouetz, line 1: I've noticed we often `pip install wheel` before installing modules of the requirements.txt file. Seems it was done for performance purpose. As wheel is mainly used it's often in the modules dependencies, so I put it first just in case. Not sure there is a real gain though

circleci/Dockerfile
```diff
@@ -0,0 +1,77 @@
...
+        python3-distutils \
+        python3-pip \
+        python3-setuptools \
+        python3-yaml \
+        snmp-mibs-downloader \
+        ssh \
+        xz-utils
+
+# Golang
+ENV GO_VERSION 1.19.9
```
chouetz, line 41: Took the opportunity to remove usage of gimme for coherence with this [PR](https://github.com/DataDog/datadog-agent-buildimages/pull/370)

circleci/requirements.txt
KSerrania: Could we pin the `wheel` version used, for safety?

.gitlab-ci.yml
```diff
@@ -204,6 +204,21 @@ build_kernel_version_testing_arm64:
...
+  except: [ tags ]
+  variables:
+    DOCKERFILE: circleci/Dockerfile
+    IMAGE: circleci-runner
+  script:
+    # Dockerhub login
+    - DOCKER_REGISTRY_LOGIN=$(aws ssm get-parameter --region us-east-1 --name ci.datadog-agent.docker_hub_login --with-decryption --query "Parameter.Value" --out text)
+    - aws ssm get-parameter --region us-east-1 --name ci.datadog-agent.docker_hub_pwd --with-decryption --query "Parameter.Value" --out text | docker login --username "$DOCKER_REGISTRY_LOGIN" --password-stdin docker.io
+    # Build
+    - docker build --build-arg BASE_IMAGE=$BASE_IMAGE --build-arg DD_TARGET_ARCH=$DD_TARGET_ARCH $CUSTOM_BUILD_ARGS --tag datadog/agent-buildimages-$IMAGE:v$CI_PIPELINE_ID-${CI_COMMIT_SHA:0:7}${ECR_TEST_ONLY} --file $DOCKERFILE .
```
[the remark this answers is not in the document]
KSerrania: I have a slight preference for handling this the same way other images are done.

But, as I understand it, this would require creating a new repository in dockerhub for each existing repository, which is probably too much effort for what it's worth. Let's just document the difference in the Go upgrade docs.

KSerrania (commented):
Could you update the PR description?

.gitlab-ci.yml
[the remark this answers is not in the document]
chouetz: I've updated the Go upgrade documentation, even if here, I could have just added a second docker repo as we have only one CircleCI image.

circleci/Dockerfile
```diff
@@ -0,0 +1,89 @@
...
+ENV CMAKE_DEST_DIR /cmake
+ENV PATH $CMAKE_DEST_DIR/bin/:$PATH
+RUN set -ex \
+    && curl -sL -O https://github.com/Kitware/CMake/releases/download/v3.13.3/$CMAKE_ARCHIVE \
+    && tar xzf $CMAKE_ARCHIVE \
+    && mv $CMAKE_NAME $CMAKE_DEST_DIR \
+    && rm $CMAKE_ARCHIVE
+
+# Codecov
+ENV CODECOV_VERSION=0.4.1
```
KSerrania, line 63: Note for later: this isn't the latest anymore (v0.6.1 is).

KSerrania (approved):

PR #394, opened 2023-05-31
Increase retry count to fix windows root certificate retrieval

KevinFairise2:
`certutil.exe -generateSSTFromWU $sstFile` is flaky and sometime fails to update certificate from Windows Update. It is likely to be a network problem. 

Increasing retry count and sleeping time fixes the problem.  It will retry every minute during a maximum of 10 minutes, in most cases the problem is fixed after 3 or 4 minutes.

### DIFF
--- a/windows/helpers/update_root_certs.ps1
+++ b/windows/helpers/update_root_certs.ps1
@@ -27,7 +27,7 @@ function Import-SSTFromWU {
     # Serialized Certificate Store File
     $sstFile = "$env:TEMP\roots.sst"
     # Generate SST from Windows Update
-    $result = Invoke-WithRetry { certutil.exe -generateSSTFromWU $sstFile } {$LASTEXITCODE -eq 0} 10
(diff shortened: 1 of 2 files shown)


windows/helpers/update_root_certs.ps1
```diff
@@ -8,14 +8,15 @@ function Invoke-WithRetry {
         Runs $command block until $BreakCondition or $RetryCount is reached.
      #>
 
-     param([ScriptBlock]$Command, [ScriptBlock] $BreakCondition, [int] $RetryCount=5, [int] $Sleep=10)
+     param([ScriptBlock]$Command, [ScriptBlock] $BreakCondition, [int] $RetryCount=10, [int] $Sleep=60)
```
bkabrda, line 11: I'd propose reducing the sleep interval and bumping retry count, e.g. sleep 30 and retry count 20. It doesn't cost us anything to do more retries, but making the build faster by potentially succeeding 30 seconds earlier would be great.

KevinFairise2: Yes I do not have any problem with this! I'll make the change

julien-lebot (approved):

PR #398, opened 2023-06-05
Clone proper branch when doing modification on version branches

KevinFairise2:


.gitlab-ci.yml
```diff
@@ -14,12 +15,12 @@ variables:
   # the buildimage for the highest Windows version supported.
   # This image must use the same Windows version as the Windows version of the Gitlab runner used in .winrelease
   WINDOWS_RELEASE_IMAGE: 486234852809.dkr.ecr.us-east-1.amazonaws.com/ci/datadog-agent-buildimages/windows_ltsc2022_x64
+  SETUP_IMAGE_NAME: 486234852809.dkr.ecr.us-east-1.amazonaws.com/ci/datadog-agent-buildimages/deb_x64:v16026304-782441d # Image used during setup task, must contains pyinvoke
```
chouetz, line 18: Do we really need to depend on a `agent-buildimage` image? I guess here the only dependency you need is `invoke`, right? Or maybe we could build this image (with just python + invoke) in buildimage and have something "dedicated" wdyt?

.gitlab-ci.yml
```diff
@@ -49,6 +50,20 @@ default:
       - unknown_failure
       - api_failure
 
+
+get_agent_version:
+  stage: setup
+  image: $SETUP_IMAGE_NAME
+  script:
+    - export VERSION=$(inv -e agent.version)
```
chouetz, line 58: not sure the `-e` brings something here (I checked if it could have saved from explicitly using `print` in your task but I guess no)

chouetz (approved):
Sounds good to me, I just gave some suggestions, I'm not sure they are valid.

PR #403, opened 2023-06-15
Re-add entrypoint for local usage of images

KSerrania:
In #370, we removed entrypoints in all build images, as they cannot be used in our new CI infrastructure.

They are still useful for developers that want to run the image locally though, therefore this PR reintroduces their usage (without the gimme setup, which isn't needed anymore as we don't use gimme).


chouquette:
Tested with this pipeline: https://gitlab.ddbuild.io/DataDog/datadog-agent/-/pipelines/19522837

chouetz (approved):

PR #405, opened 2023-06-19
CI(Notification): Add link to the failed pipeline in the Slack message

KevinFairise2:
It is more convenient to have a direct link to the pipeline in the message


amenasria (approved):
LGTM

PR #402, opened 2023-06-13
Upgrade to go1.19.10

ogaca-dd:
Upgrade to go1.19.10

PR #409, opened 2023-06-27
Revert "Update Go version to 1.20.5 (#408)"

Kaderinho:
Go upgrade is causing issue on the Agent side (go tests are taking way longer) and the version mismatch between this repo and the Agent repo is blocking 7.47 preparation


julien-lebot (approved):

KSerrania:
From what I'm seeing on `main` of `datadog-agent` ([here](https://github.com/DataDog/datadog-agent/blob/main/.go-version)), the go 1.19.10 upgrade (from https://github.com/DataDog/datadog-agent-buildimages/pull/402) was only done in `7.46.x` of `datadog-agent`, and not in `main` (https://github.com/DataDog/datadog-agent/pull/17619), therefore even with this revert, `datadog-agent-buildimages` and `datadog-agent` are in incompatible states (which prevents using new images in `datadog-agent`). I see two options:
- also reverting https://github.com/DataDog/datadog-agent-buildimages/pull/402,
- porting https://github.com/DataDog/datadog-agent/pull/17619 to `main`.

KSerrania:
> I think 1.19.10 is desirable in any case so my guess is that we should port https://github.com/DataDog/datadog-agent/pull/17619 to main imho

👍 In that case, let's merge this PR, and update `main` of `datadog-agent` with the image from this PR's merge + the port of https://github.com/DataDog/datadog-agent/pull/17619.

KSerrania (approved):

PR #420, opened 2023-07-24
Use a nightly version of Rust on RPM

alopezz:
Required by https://github.com/DataDog/datadog-agent/pull/18303.

The current rpm image contains an old version of glibc for compatibility reasons, and that makes it harder to support newer dependencies on Agent integrations. We're targeting `musl` for Rust-based extensions instead for this image as a way to work around that.

First attempt of https://github.com/DataDog/datadog-agent/pull/18303 had this error, which this PR addresses:

```
        = note: the `x86_64-unknown-linux-musl` target may not be installed
        = help: consider downloading the target with `rustup target add x86_64-unknown-linux-musl
```

PR #421, opened 2023-07-25
Bump rust and rustup versions

vivek-datadog:
Motivation: https://github.com/DataDog/datadog-agent-buildimages/pull/420#issuecomment-1648974745


FlorentClarret:
Fixed in https://github.com/DataDog/datadog-agent-buildimages/pull/420

PR #424, opened 2023-07-27
Bump more reference of invoke

chouetz:
I've discovered we were using 2 other reference of invoke (in addition to the requirement.txt file)
Trying to bump these references as well for datadog-agent


spencergilbert (commented):
Failed `test_rpm_x64`

chouetz:
Image tested successfully with this [pipeline](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/pipelines/18573762)

PR #426, opened 2023-07-28
Back to standard clang-format installation

chouetz:
Patch [done](https://github.com/llvm/llvm-project/issues/64120) on LLVM side , we can use back the standard (latest) version


chouetz:
Waiting for `build_circleci_runner` step to be OK before any action on this PR

chouquette (approved):

PR #433, opened 2023-08-15
Add yq to deb x86 runner

akarpz:
Adds `yq` to the deb-x64 dockerfile so it is available when the updated patch script for setting the image tag on agent staging clusters is called: https://github.com/DataDog/k8s-datadog-agent-ops/pull/2253


chouquette (commented):
Could you also add a checksum to ensure the binary is what's expected?

Otherwise LGTM :)

PR #435, opened 2023-08-22
build-gcc: retry downloading gcc if it fails

chouquette:


spencergilbert (approved):

chouquette:
closing as the curlrc file that was added since this PR was opened achieve the same goal in a much cleaner way

PR #438, opened 2023-09-06
Factorize requirements.txt files

chouetz:
Use reference to files for requirements.txt:
* prevent ci-uploader from having its own definition
* allows possibility to target these references from datadog-agent, so globally reduces the risk to have duplication in python module and versions

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,18 +1,15 @@
 # python development requirements for the Datadog Agent
-invoke==2.2.0
 reno==3.5.0
 docker==6.0.1; python_version >= '3.7'
 docker==5.0.3; python_version < '3.7'
 docker-squash==1.0.9
 dulwich==0.20.45
-requests==2.27.1
 toml==0.10.2
 packaging==21.3
 # more recent boto3 has a bug which generates an import exception.
 # pin until this can be resolved.
 boto3==1.22.4
-botocore==1.25.4 ## for awscli
-awscli==1.23.4
+-r https://raw.githubusercontent.com/DataDog/datadog-agent-buildimages/nschweitzer/requirements/requirements/common.txt
 # urllib3 major version 2, released on May 4th 2023, breaks botocore used
 # by awscli (removed DEFAULT_CIPHERS list from urllib3.util.ssl_)
 urllib3==1.26.15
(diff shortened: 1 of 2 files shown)


datadog-ci-uploader/requirements.txt
```diff
@@ -0,0 +1,4 @@
+-r https://raw.githubusercontent.com/DataDog/datadog-agent-buildimages/nschweitzer/requirements/requirements/requests_awscli.txt
```
chouetz, line 1:
```suggestion
-r https://raw.githubusercontent.com/DataDog/datadog-agent-buildimages/main/requirements/requests_awscli.txt
```
Will will need to do this afterwards. We might want to pin it to a certain branch version to ensure reproducibility in datadog-agent (see how it's [referenced](https://github.com/DataDog/datadog-agent/blob/main/requirements.txt#L2) today)

chouquette: Wouldn't it be simpler to split the addition of this file in a dedicated PR, merge it, and simply reference it from main afterward?

spencergilbert: Agreed, either way works - but @chouquette's proposal seems a little bit cleaner.

KevinFairise2:
Do you think it would be possible to have only on `requirements.txt` and extract the dependencies we need with something like `grep` (or something else). Like we do here https://github.com/DataDog/test-infra-definitions/blob/main/Dockerfile#L98-L99.
Because I feel like we could end with a lot of different `requirements.txt` files with only on dependency to match all the possible requirements for our Docker images

requirements.txt
```diff
@@ -1,18 +1,15 @@
...
 dulwich==0.20.45
-requests==2.27.1
 toml==0.10.2
 packaging==21.3
 # more recent boto3 has a bug which generates an import exception.
 # pin until this can be resolved.
 boto3==1.22.4
-botocore==1.25.4 ## for awscli
-awscli==1.23.4
+-r https://raw.githubusercontent.com/DataDog/datadog-agent-buildimages/nschweitzer/requirements/requirements/common.txt
```
chouquette, line 12: This should be
```suggestion
-r https://raw.githubusercontent.com/DataDog/datadog-agent-buildimages/main/requirements/common.txt
```
no?

spencergilbert (approved):

PR #440, opened 2023-09-07
[install_java.ps1] Fix bad sha variable

pgimalac:
I think the wrong variable is used in `install_java.ps1`


julien-lebot (approved):
Nice catch

PR #443, opened 2023-09-08
[ASCII-731] Add invoke task to update go version

pgimalac:
Create an invoke task to automatically fetch the SHA256 of all needed Go archives and update every `GO_VERSION` and `GO_SHA`.

The SHA256 are collected directly from https://storage.googleapis.com/golang (which is an official source).

I added an optional `check_archive` argument which will also fetch the archives from https://go.dev/dl and compare their SHA256 to the expected ones.


tasks/update_go.py
```diff
@@ -0,0 +1,171 @@
...
+    "./rpm-arm64/Dockerfile": 2,
+    "./rpm-armhf/Dockerfile": 2,
+    "./rpm-x64/Dockerfile": 2,
+    "./suse-x64/Dockerfile": 2,
+    "./system-probe_arm64/Dockerfile": 2,
+    "./system-probe_x64/Dockerfile": 2,
+}
+
+
+# returns a map from a pattern to what it should be replaced with, for dockerfiles
```
chouetz, line 38: you could put this as the method's docstring, wdyt? (same comment applies for other methods)

tasks/update_go.py
```diff
@@ -0,0 +1,171 @@
...
+        pattern = re.compile(f"^(ARG {varname}=)[a-z0-9]+$", flags=re.MULTILINE)
+        replace = rf"\g<1>{sha}"
+
+        patterns[pattern] = replace
+
+    return patterns
+
+
+# returns a map from a pattern to what it should be replaced with, for windows file
+def _get_windows_patterns(
```
chouetz, line 58: could this method and the previous one share their code? There are very few differences.

tasks/update_go.py
```diff
@@ -0,0 +1,171 @@
...
+
+# checks that the archive sha is the same as the given one
+def _check_archive(version: str, shas: Dict[Platform, str]):
+    for (os, arch), expected_sha in shas.items():
+        ext = _get_archive_extension(os)
+        url = f"https://go.dev/dl/go{version}.{os}-{arch}.{ext}"
+        print(f"[check-archive] Fetching archive at {url}", file=sys.stderr)
+        # Using `curl` through `ctx.run` takes way too much time due to the archive being huge
+        # use `requests` as a workaround
+        req = requests.get(url)
```
chouetz, line 111: You should probably check the `req.status_code`? Otherwise there's a risk to crash on next line.

tasks/update_go.py
```diff
@@ -0,0 +1,171 @@
...
+        sha = hashlib.sha256(req.content).hexdigest()
+        if sha != expected_sha:
+            raise exceptions.Exit(
+                f"The SHA256 of Go on {os}/{arch} should be {expected_sha}, but got {sha}"
+            )
+
+
+# replace patterns in a file
+def _handle_file(path: str, patterns: Dict[re.Pattern, str], expected_match: int = 1):
```
chouetz, line 120: hopefully we will have a possibility to share this within a module between repositories soon

circleci/Dockerfile
```diff
@@ -38,6 +38,7 @@ RUN set -ex \
     xz-utils
 
 # Golang
+# The format of the GO args shouldn't be changed as the update-go invoke task relies on it
```
chouetz, line 41:
```suggestion
# If you want to change the 2 below lines, please use the update-go invoke task
```
I suggest this because the other way around could be misleading, wdyt?

brycekahle (approved):

tasks/update_go.py
pgimalac: They were similar because I copy pasted the first to write the second :smile:
I simplified `_get_windows_pattern` to just return the patterns really used in the windows file

tasks/update_go.py
```diff
@@ -0,0 +1,152 @@
...
+        patterns[pattern] = replace
+
+    return patterns
+
+
+def _get_windows_patterns(version: str, shas: Dict[Platform, str]) -> Dict[re.Pattern, str]:
+    """returns a map from a pattern to what it should be replaced with, for windows file"""
+    sha = shas[("windows", "amd64")]
+    return {
+        re.compile('^(    "GO_VERSION"=")[.0-9]+(";)$', flags=re.MULTILINE): rf"\g<1>{version}\g<2>",
```
julien-lebot, line 57: Do we need the exact number of spaces, instead of using e.g. `\s+` ?

pgimalac: Nice catch, I initially didn't use capture groups (`\g<1>`) so I didn't have a way to keep the same number of spaces, I'll update that !

julien-lebot (approved):

ogaca-dd (approved):

PR #444, opened 2023-09-08
CI: fix setup-env error detection

chouquette:
When using `export FOO=$(something)` the overall command will always return a success code, even if the subshell failed. In this case since we don't need to export the variable anyway, we can remove it and properly detect failures


chouquette:
Problematic build: https://gitlab.ddbuild.io/DataDog/datadog-agent-buildimages/-/jobs/325569297
Test build with the fix: https://gitlab.ddbuild.io/DataDog/datadog-agent-buildimages/-/jobs/325609230

PR #449, opened 2023-09-14
Add a new job to warn user when his images are ready to be used

chouetz:
Goal is to be alerted when the images are ready for a test in `datadog-agent`
here is the new message
image

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,3 +1,5 @@
+include: "https://gitlab-templates.ddbuild.io/slack-notifier/sdm/template.yml"
+
 stages:
   - setup
   - build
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -536,6 +534,23 @@ release_circleci_runner:
   after_script:
     - docker rmi datadog/agent-buildimages-$IMAGE:v$CI_PIPELINE_ID-${CI_COMMIT_SHA:0:7} datadog/agent-buildimages-$IMAGE:latest
 
+.slack-notifier-base:
+  stage: notify
```
KevinFairise2, line 538: Maybe it would be clearer to not indicate the stage here. Since this one is extended by only two jobs that have different stages (notify, availability_image). WDYT?

KevinFairise2 (approved):
LGTM, just left a minor suggestion. 🚀 
The `/usr/local/bin/notify.sh` script is defined in the slack-notifier image, right?

.gitlab-ci.yml
chouetz: I will use the template provider by `#ci-cd` then I will have to overwrite it only for the additional stage

chouetz:
> LGTM, just left a minor suggestion. 🚀 The `/usr/local/bin/notify.sh` script is defined in the slack-notifier image, right?

yes the script is in the slack-notifier image

PR #450, opened 2023-09-21
[pip] add requirements for e2e tests

pducolin:
E2E tests run on an image built in test-infra-definitions and used in datadog-agent with invoke tasks from datadog-agent. Some of the datadog-agent build image are skipped as we have aws-cli and other dependencies coming from Pulumi


vboulineau (approved):

requirements/e2e.txt
chouetz, line 1: reno, docker, toml, and packaging could be grouped as they are both requirements for e2e and main requirements.txt, correct?

pducolin: I hesitate grouping them, they are not logically related, they are grouped by chance as they are accidentally all required by e2e and other agent tasks. I prefer keeping them separated, this fixes the duplication issue as we still have one single spot where version is defined

pducolin:
Merge queue timed out but passed https://gitlab.ddbuild.io/DataDog/datadog-agent-buildimages/-/pipelines/20369612

PR #453, opened 2023-09-25
[CI] Add `constraints.txt` and unnest the `requirements.txt` files

amenasria:
The nested `requirements.txt` file are really hard to read and not easily maintanable.

I propose instead using [a constraints file](https://pip.pypa.io/en/stable/user_guide/#constraints-files) which is way easier to maintain and which centralize dependencies.

PR #457, opened 2023-09-28
Move comment to `constraints.txt`

amenasria:
Small miss of #453

### DIFF
--- a/requirements/constraints.txt
+++ b/requirements/constraints.txt
@@ -1,7 +1,5 @@
 awscli==1.29.45
 black==22.10.0
-# urllib3 major version 2, released on May 4th 2023, breaks botocore used
-# by awscli (removed DEFAULT_CIPHERS list from urllib3.util.ssl_)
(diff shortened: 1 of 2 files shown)


requirements/constraints.txt
```diff
@@ -1,5 +1,7 @@
 awscli==1.29.45
 black==22.10.0
+# urllib3 major version 2, released on May 4th 2023, breaks botocore used
+# by awscli (removed DEFAULT_CIPHERS list from urllib3.util.ssl_)
```
chouetz, line 4: should the comment be above urllib3 instead?

amenasria: This comment justifies the version pinning, so it should be above the version pinning hence in `constraints.txt` IMO. WDYT?

amenasria: Oops mb got you ! Fixed in 9fa38ba

chouetz (approved):

PR #464, opened 2023-10-20
Set default configurations for auto-retry

chouetz:
Full [pipeline](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/pipelines/22217903) triggered from this [branch](https://github.com/DataDog/datadog-agent/compare/main...nschweitzer/inv_buildimages)

### DIFF
--- a/circleci/Dockerfile
+++ b/circleci/Dockerfile
@@ -88,5 +88,9 @@ COPY requirements.txt /
 COPY requirements /requirements
 RUN python3 -m pip install -r requirements.txt -r requirements/circleci.txt
 
+# External calls configuration
+COPY .awsconfig $HOME/.aws/config
+COPY .curlrc .wgetrc $HOME/
+
 # Setup entrypoint
 WORKDIR $GOPATH
--- a/.awsconfig
+++ b/.awsconfig
@@ -1,0 +1,3 @@
+[default]
+retry_mode = standard
+max_attempts = 5
--- a/.curlrc
+++ b/.curlrc
@@ -1,0 +1,6 @@
+--continue-at -
+--include
+--location
+--retry 10
+--max-time 120
+--silent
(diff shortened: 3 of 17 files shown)


brycekahle (approved):

circleci/Dockerfile
```diff
@@ -88,5 +88,9 @@ COPY requirements.txt /
 COPY requirements /requirements
 RUN python3 -m pip install -r requirements.txt -r requirements/circleci.txt
 
+# External calls configuration
+COPY .awsconfig .aws/config
+COPY .curlrc .wgetrc ./
```
chouquette, line 93: Shouldn't the destination be more explicit? In case we change the `WORKDIR` beforehand for whatever reason

windows/Dockerfile
```diff
@@ -63,6 +63,10 @@ COPY python-packages-versions.txt requirements.txt requirements-py2.txt /
 COPY requirements /requirements
 COPY ./windows/set_cpython_compiler.cmd set_cpython_compiler.cmd
 
+# External calls configuration
+COPY .awsconfig .aws/config
+COPY .curlrc .wgetrc ./
```
julien-lebot, line 68: That's not used on Windows, is it ?
https://github.com/DataDog/datadog-agent-buildimages/blob/main/windows/helpers.ps1#L64

amenasria (approved):
LGTM ! Nice addition !

We should check that every `curl` occurence in the datadog-agent repository is not impacted by this though !

windows/Dockerfile
julien-lebot: I didn't find any references to curl in the Windows folder - maybe auto-retry can be added in the helper PowerShell directly ?
In any case maybe you can remove it for this PR.

PR #472, opened 2023-11-02
CircleCI Image migration to gcr

Pythyu:
The `circleci_runner` is now deployed on gcr as well since we're switching to free dockerhub.

To gather the status of the build on GCR, we're using downstream child pipeline. To do so, the docker tag has been slightly edited from `${CI_COMMIT_SHA:0:7}` to `${CI_COMMIT_SHORT_SHA}` on the circleci image since trigger doesn't handle computation.


chouquette (commented):
There are other occurences of `CI_COMMIT_SHA`, don't we want to update those as well?

chouquette (approved):
I'd feel more at ease if we were to unify the build images short sha computation, but this LGTM as is nonetheless :)

Pythyu:
Yea good point was planning to do it in a separate PR but I'll do it here should work all the same

chouquette (approved):

PR #474, opened 2023-11-07
Don't build go from source deb-x64

amenasria:
Since #160 we've been building go from source for the deb-x64 image. Since we're now using Ubuntu 14.04, let's try to see if we can install it directly.


deb-x64/Dockerfile
```diff
@@ -156,15 +157,10 @@ RUN curl -Sl -O https://dd-agent-omnibus.s3.amazonaws.com/kernel-4.9-headers-deb
...
 
 # Go
-# Starting with go1.16, we need to build go from source, as the publicly distributed
-# go binaries are built with a linker (binutils) version that's higher than what we have in
-# the deb_x64 builder (the builder has 2.22, while go1.16 uses 2.26). Therefore, we need to
-# build go manually in the builder for it to work with the linker present in the builder image.
-# See: https://github.com/golang/go/issues/43996#issuecomment-782547917
-# TODO: investigate if we could use gvm to do this for us.
-COPY setup_go.sh /
-RUN ./setup_go.sh
```
chouquette, line 166: Do we still need bison environment variables in the `Dockerfile` now that this script isn't executed?

I think `binutils` is still required when building gcc though

deb-x64/Dockerfile
```diff
@@ -156,15 +157,10 @@ RUN curl -Sl -O https://dd-agent-omnibus.s3.amazonaws.com/kernel-4.9-headers-deb
     && rm kernel-4.9-headers-deb-x64.tgz
 
 # Go
-# Starting with go1.16, we need to build go from source, as the publicly distributed
-# go binaries are built with a linker (binutils) version that's higher than what we have in
-# the deb_x64 builder (the builder has 2.22, while go1.16 uses 2.26). Therefore, we need to
-# build go manually in the builder for it to work with the linker present in the builder image.
-# See: https://github.com/golang/go/issues/43996#issuecomment-782547917
-# TODO: investigate if we could use gvm to do this for us.
-COPY setup_go.sh /
```
chouquette, line 165: As mentionned on slack, I'd be in favor of removing the script since it isn't used anymore. If we ever need it we'll still need to revert this commit

amenasria: That's actually a good point, I realized that removing it in the same PR and the squashing would make it easier for people to find it back !

deb-x64/Dockerfile
amenasria: Right, I'll remove them, first checking if it works (retriggering the CI)

PR #476, opened 2023-11-09
[Pipeline] Replace tests by running a datadog-agent child pipeline

Pythyu:
Since [we’re building system probe in omnibus](https://github.com/DataDog/datadog-agent/pull/19492), datadog-agent-buildimages CI is failing because tests are trying to build the agent without building system-probe, which is now required. Hence we're replacing tests by a datadog-agent child pipeline to fix current and future issues about how the agent is builded.
Full details on [this document](https://docs.google.com/document/d/1udjaYBdw82ZV7aVsJ9ZeFPVzlcDLLgY9c2xr5Zo1Jag/edit).

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -466,55 +466,55 @@ build_windows_ltsc2022_x64:
 
 release_deb_x64:
   extends: .release
-  needs: ["test_deb_x64"]
+  needs: ["trigger_tests"]
   variables:
     IMAGE: deb_x64
 
 release_rpm_x64:
   extends: .release
-  needs: ["test_rpm_x64"]
+  needs: ["trigger_tests"]
   variables:
     IMAGE: rpm_x64
 
 release_suse_x64:
   extends: .release
-  needs: ["test_suse_x64"]
+  needs: ["trigger_tests"]
   variables:
     IMAGE: suse_x64
 
 release_deb_arm64:
   extends: .release
-  needs: ["test_deb_arm64"]
+  needs: ["trigger_tests"]
   variables:
     IMAGE: deb_arm64
 
 release_rpm_arm64:
   extends: .release
-  needs: ["test_rpm_arm64"]
+  needs: ["trigger_tests"]
   variables:
     IMAGE: rpm_arm64
 
 release_system-probe_x64:
   extends: .release
-  needs: ["test_system-probe_x64"]
+  needs: ["trigger_tests"]
   variables:
     IMAGE: system-probe_x64
 
 release_system-probe_arm64:
   extends: .release
-  needs: ["test_system-probe_arm64"]
+  needs: ["trigger_tests"]
   variables:
     IMAGE: system-probe_arm64
 
 release_datadog-ci-uploader:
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -260,42 +278,6 @@ trigger_build_kernels:
...
-  needs: ["get_agent_version", "build_rpm_arm64"]
-  tags: ["runner:docker-arm", "platform:arm64"]
-
-test_suse_x64:
-  extends: .test_agent6
-  image: 486234852809.dkr.ecr.us-east-1.amazonaws.com/ci/datadog-agent-buildimages/suse_x64${ECR_TEST_ONLY}:v$CI_PIPELINE_ID
-  needs: ["get_agent_version", "build_suse_x64"]
-  tags: ["runner:main"]
-
 .test_system_probe:
```
chouquette, line 281: This isn't used anymore no?

.gitlab-ci.yml
```diff
@@ -455,63 +405,63 @@ test_windows_ltsc2022_x64:
 
 release_deb_x64:
   extends: .release
-  needs: ["test_deb_x64"]
+  needs: ["trigger_tests"]
```
chouquette, line 408: I think you can move the `needs` into the `.release` section, which would reduce the duplicated boilerplate

.gitlab-ci.yml
```diff
@@ -243,6 +243,24 @@ trigger_build_kernels:
       - local: .gitlab/kernel_version_testing.yml
     strategy: depend
 
+trigger_tests:
+  stage: test
+  except: [tags, schedules]
+  variables:
+    RUN_KITCHEN_TESTS: "false"
```
chouquette, line 250: Shouldn't we run all kitchen tests for this kind of changes as well? It seems similar in term of risks to changes in omnibus-software

chouquette:
Besides the few questions/remarks, LGTM on principle :)

KevinFairise2:
I have a small question, it looks like the triggered pipeline does not appear in the list of the "Run checks", do we have a way to display them here? I am afraid that we will never know when the triggered pipeline fails and we will always ignore it.

chouquette:
> I have a small question, it looks like the triggered pipeline does not appear in the list of the "Run checks", do we have a way to display them here? I am afraid that we will never know when the triggered pipeline fails and we will always ignore it.

In order to be displayed there, a job *must* be a non trigger job, ie. it must have a `script` section. Its name must also not match one of these patterns: `notify,merge,ng:build,ng:debs,tag,debs,ci-image,after-build,after-test`

.gitlab-ci.yml
Pythyu: previous tests were only testing to build the agent, I'll rather stick to it to not increase to much CI time

amenasria:
**👏 praise:** ‏Can't wait for this to be merged 🙀

chouquette (approved):
LGTM, having a non-trigger job so we can mark it as required in github can come at a later point in order to unblock people requiring new buildimages

PR #482, opened 2023-11-27
debian: install pxz

chouquette:
PXZ is a parallel compressor for the XZ format. It will allow us to create smaller gomods archive with no time penalty


amenasria (approved):
LGTM

PR #485, opened 2023-11-30
Add patchelf to our linux images

chouquette:
This allows us to modify the rpath for some 3rd parties binary artifacts, which we can't build with the correct rpath from the start.


KevinFairise2 (approved):
LGTM. However I am not sure to understand why you add it to `deb-arm` and not to `deb-x64`?

chouquette:
> LGTM. However I am not sure to understand why you add it to `deb-arm` and not to `deb-x64`?

Hm, that looks like a mistake somewhere, and that doesn't really reflect with the tests that were done, I need to look more into it... 
Thanks for noticing!

chouquette:
> > LGTM. However I am not sure to understand why you add it to `deb-arm` and not to `deb-x64`?
> 
> Hm, that looks like a mistake somewhere, and that doesn't really reflect with the tests that were done, I need to look more into it... Thanks for noticing!

Oh well I just didn't remember that it was already added to the x86 images :sweat_smile: I'll reword the commits

PR #491, opened 2023-12-05
Add rust to windows build container

ISauve:
New project requires rust to be added to the windows build container.


windows/helpers/phase3/install_rust.ps1
```diff
@@ -0,0 +1,23 @@
...
+$ErrorActionPreference = 'Stop'
+
+$isInstalled, $isCurrent = Get-InstallUpgradeStatus -Component "rust" -Keyname "version" -TargetValue $Version
+if($isInstalled -and $isCurrent) {
+    Write-Host -ForegroundColor Green "Rust up to date"
+    return
+}
+# No need to check for previous installations; rust-up will remove them.
+
+$source="https://static.rust-lang.org/rustup/dist/x86_64-pc-windows-msvc/rustup-init.exe"
```
ISauve, line 15: Unfortunately I wasn't able to find a versioned rust install link. This means that when a new rust version comes out, the windows image build will likely start failing until the  `RUST_SHA256` value is updated.
I figured this was preferable to not checking the hash at all, but it's definitely not ideal.

julien-lebot: Rustup is a tool that manages multiple Rust toolchain.
When a new version of Rust comes out the `rustup` installer isn't likely to change.
Conversely, an update to the `rustup` tool could happen without a new version of Rust coming out.

I think it's fine to check the SHA256 of the `rustup` tool, but to update Rust in our image we should use a different mechanism, such as those described in https://rust-lang.github.io/rustup/overrides.html

windows/helpers/phase3/install_rust.ps1
```diff
@@ -0,0 +1,23 @@
...
+
+Write-Host -ForegroundColor Green "Installing rust-up"
+$out = "$($PSScriptRoot)\rustup-init.exe"
+Get-RemoteFile -RemoteFile $source -LocalFile $out -VerifyHash $Sha256
+
+& $out -y
```
julien-lebot, line 21: The $Version variable is not used here.
```suggestion
& $out -y --default-toolchain none
Reload-Path
rustup override set $Version
```

ISauve:
@julien-lebot My apologies for using terms interchangeably and added some confusion: the version pinned is the `rust-up` version, not the rust language version. Currently the newest version of `rust-up` is 1.26.0, and it installs rustc `1.74.0`. Since using `rust-up` is the preferred install mechanism, is significantly easier than installing rust directly, and ensures that we install all the rust components (ie `cargo`, which we need, & I'm not sure if the direct rust installation also installs that) - I would lean towards using `rust-up` rather than using the chocolately package.

ISauve:
The caveat to using `rust-up` is that the install link for `rust-up` is not versioned, so (a) the `$version` env variable doesn't actually do anything other than provide documentation re: which version of rust-up the `sha256` refers to, and (b) when a new `rust-up` version is released, that's when the build image will start failing. I'm not sure how often `rust-up` is updated, so this may not be a big issue, but it's something to note.

FWIW, the project has no dependency on a specific version of rust, so I was fine to let `rust-up` just install the latest, but we can certainly add a flag to pin a specific `rust` version if you think that's necessary.

windows/helpers/phase3/install_rust.ps1
ISauve: I'm resolving this thread as I've updated the naming & fixed the install link - let's continue the discussion re: install mechanism in the PR below :)

spencergilbert (approved):

julien-lebot (approved):

PR #495, opened 2023-12-14
[DRAFT] Improve the go version update task with pr creation

jeremy-hanna:
This is an attempt to script some of the annoying parts re: Go version update detailed in the [RFC](https://docs.google.com/document/d/1X7T3X5_mMRb7pd4zTQ-JluV_5UUxRk-IYOgMoV3S7Ug/edit#heading=h.ixldy9ty2xy5)
* adds `wait_for_test` branch name which will be created in `datadog-agent` after we get the CI_PIPELINE_ID
* runs the current invoke task to update the Dockerfiles
* creates a `datadog-agent-buildimages` PR (to start the CI pipeline)
* gets the CI_PIPELINE_ID from gitlab via it's api to determine the docker image tag the `datadog-agent` CI needs
* runs the `datadog-agent` invoke task with the Go version and image tag from above 
* creates the `datadog-agent` branch with correct CI image tag and Go versions (the `datadog-agent` CI will fail because the images won't be built yet, but that is fine since the `wait_for_test` job will verify the upgrade in the buildimage repo's CI)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,5 +12,6 @@ invoke
 packaging
 reno
 requests
+semver==2.10.0
 toml
 urllib3
--- a/tasks/update_go.py
+++ b/tasks/update_go.py
@@ -162,7 +162,7 @@ def _update_and_create_pr(gover):
         if local_uncommited_changes_exist(repo):
             raise exceptions.Exit("Uncommited changes exist in datadog-agent-buildimages repo. Exiting.")
         checkout_latest_main(repo)
-        _add_wait_for_tests(f"update-go-{gover}")
+        _add_wait_for_tests(branch)
         _update_images(gover)
         if local_uncommited_changes_exist(repo):
             create_branch_and_push_changes(
@@ -198,7 +198,7 @@ def _update_and_create_pr(gover):
         res = requests.get(f"https://gitlab.ddbuild.io/api/v4/projects/291/jobs/{build_ids[0]}", headers={"PRIVATE-TOKEN": os.environ["GITLAB_ACCESS_TOKEN"]})
         pipeline_id = res.json()["pipeline"]["id"]
         image_tag = f"v{pipeline_id}-{commit_sha[:8]}_test_only"
-        _create_agent_pr(gover, image_tag)
+        _create_agent_pr(branch, gover, image_tag)
 
     create_pr_link = f"https://github.com/DataDog/datadog-agent-buildimages/compare/{branch}?expand=1"
     print(f"Opening link to create an associated PR for the updated buildimages: {create_pr_link}")
@@ -209,7 +209,7 @@ def _update_and_create_pr(gover):
     subprocess.run(["open", create_pr_link])
 
 
-def _create_agent_pr(gover, image_tag):
+def _create_agent_pr(branch, gover, image_tag):
     repo = "datadog-agent"
     with dd_repo_temp_cwd(repo):
         if local_uncommited_changes_exist(repo):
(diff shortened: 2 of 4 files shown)


requirements.txt
```diff
@@ -12,5 +12,6 @@ invoke
 packaging
 reno
 requests
+semver==2.10.0
```
jeremy-hanna, line 15: this is needed by the `datadog-agent`'s update-go invoke task

tasks/update_go.py
```diff
@@ -120,15 +126,106 @@ def _handle_file(path: str, patterns: Dict[re.Pattern, str], expected_match: int
...
+    with_branch = f"""{trigger}
+    branch: {branch}"""
+    with open(".gitlab-ci.yml", "r") as f:
+        contents = f.read()
+    contents = re.sub(trigger, with_branch, contents, count=1)
+    with open(".gitlab-ci.yml", "w") as f:
+        f.write(contents)
+
+# TODO: not sure how to run this before merging
+def _remove_wait_for_tests(branch: str):
```
jeremy-hanna, line 147: It would be nice to have this run with a main build so we didn't need another commit, I'm not completely sure how to do that right now but it may be possible with a label and the github api

tasks/update_go.py
```diff
@@ -120,15 +126,106 @@ def _handle_file(path: str, patterns: Dict[re.Pattern, str], expected_match: int
...
+                    continue
+                else:
+                    raise exceptions.Exit("Exhausted retries waiting for gitlab job to start.")
+            else:
+                build_ids.extend(matches)
+                print(f"Gitlab job started with build_ids: {build_ids}")
+                break
+
+        # XXX: is it OK to require the user create a gitlab access token?
+        res = requests.get(f"https://gitlab.ddbuild.io/api/v4/projects/291/jobs/{build_ids[0]}", headers={"PRIVATE-TOKEN": os.environ["GITLAB_ACCESS_TOKEN"]})
```
jeremy-hanna, line 198: it would be very nice to determine the gitlab pipeline id without needing an access_token but I just [used the gitlab api](https://docs.gitlab.com/ee/api/jobs.html#get-a-single-job)

tasks/utils.py
```diff
@@ -0,0 +1,36 @@
...
+        repo,
+        branch,
+        commit_msg,
+        commit_args= None,
+        labels= None,
+        ):
+    subprocess.check_call(["git", "checkout", "-b", branch])
+    subprocess.check_call(["git", "add", ".", ":!venv"])
+    subprocess.check_call(["git", "commit", "-am", commit_msg])
+    subprocess.check_call(["git", "push", "origin", branch])
```
pgimalac, line 36: I don't like the idea of having a script which creates branches and pushes things in my name, doing that from a workflow would make it clear it's automated (and using Github actions all the git logic disappears, you just call an action)

tasks/update_go.py
```diff
@@ -120,15 +126,106 @@ def _handle_file(path: str, patterns: Dict[re.Pattern, str], expected_match: int
...
+        if local_uncommited_changes_exist(repo):
+            raise "Uncommited changes exist in datadog-agent repo. Exiting."
+        checkout_latest_main(repo)
+        out = subprocess.check_output('inv', '-e', 'update-go', gover, image_tag)
+        if local_uncommited_changes_exist(repo):
+            create_branch_and_push_changes(
+                repo,
+                branch,
+                f"Update Go version to {gover} and buildimages to {image_tag} for datadog agent",
+            )
```
pgimalac, line 224: Having a task which cd to a different repository to run some other commands (in particular creating a PR) seems too risky to me

pgimalac (changes requested):
My thought (and what I described in the RFC) was to do this automation directly in a Github action, not in a script running locally
We can sync to talk about it

PR #496, opened 2023-12-14
Enable parallel XZ compression for deb & rpm packages [APL-2626]

chouquette:
This PR provides `dpkg` and `rpmbuild` binary build with `liblzma` 5.2, which supports parallel XZ compression.

~Please note that for ubuntu/debian, this is provided as a PPA as the `dpkg` version we need doesn't build easily on our build image (it would require bumping at least gettext & automake manually), however I understand if that's an issue~


deb-x64/Dockerfile
```diff
@@ -94,7 +97,7 @@ RUN curl -OL https://www.kernel.org/pub/software/scm/git/git-${GIT_VERSION}.tar.
     # --no-same-owner: git tarball has a file with UID 110493 which makes pulling this image fail, because we use docker user namespacing and we can't have >65K UIDs. \
     && tar xzf git-${GIT_VERSION}.tar.gz --no-same-owner \
     && cd git-${GIT_VERSION} \
-    && make prefix=/usr/local all \
+    && make -j$(nproc) prefix=/usr/local all \
```
chouetz, line 100: Nice!

chouetz (approved):

PR #497, opened 2023-12-15
Fix scheduled pipeline & remove deprecated `only`/`except` keywords

chouquette:


.gitlab-ci.yml
```diff
@@ -84,7 +99,10 @@ get_agent_version:
 
 .build:
   stage: build
-  except: [tags]
+  rules:
+    - if: $CI_COMMIT_TAG != null
+      when: never
```
Pythyu, line 104: ❓ Question :
What does `CI_COMMIT_TAG` refer to again ? does this mean we need at least a PR tag to run the CI ? or if it's a commit tag it means that a simple `git commit -m <msg> && git push` won't trigger the pipeline ?

Pythyu (approved):
Aside from my comment LGTM

.gitlab-ci.yml
```diff
@@ -216,7 +234,10 @@ build_kernel_version_testing_arm64:
 build_circleci_runner:
   extends: [.x64]
   stage: build
-  except: [tags]
+  rules:
+    - if: $CI_COMMIT_TAG != null
+      when: never
+    - when: always
```
KevinFairise2, line 240: nit: this should be equivalent
```suggestion
  rules:
    - if: $CI_COMMIT_TAG == null
```

.gitlab-ci.yml
chouquette: `The commit tag name. Available only in pipelines for tags.`
So if a pipeline is created with `git tag tagname && git push origin tagname` this job will not be added to the pipeline

KevinFairise2 (approved):
Small nit, but you can let is as is. LGTM

.gitlab-ci.yml
Pythyu: Aah right good 👍

.gitlab-ci.yml
chouquette: Indeed! Fixed

PR #498, opened 2023-12-16
[ASCII-983] Github workflow to open go update PR

pgimalac:
Add a Github workflow which can be triggered manually to open a PR to update the Go version.

This can be tested using Github CLI:
```
gh workflow -R DataDog/datadog-agent-buildimages run go-update.yml -f go_version=1.21.5 --ref pgimalac/auto-create-pr-go-update
```

Note: this is not supposed to be perfect already, more a POC.
For now it can only be triggered manually, doesn't notify on slack, etc

### DIFF
--- a/.github/workflows/go-update.yml
+++ b/.github/workflows/go-update.yml
@@ -8,7 +8,7 @@ on:
         required: true
         type: string
       open_pr:
-        description: 'Whether to a PR'
+        description: 'Whether to open a PR'
         required: false
         type: boolean
         default: true
--- a/.github/workflows/go-update.yml
+++ b/.github/workflows/go-update.yml
@@ -8,7 +8,6 @@ on:
         required: true
         type: string
         default: ''
-  pull_request:
 
 jobs:
   open-go-update-pr:
--- a/.github/workflows/go-update.yml
+++ b/.github/workflows/go-update.yml
@@ -7,11 +7,24 @@ on:
         description: 'Go version'
         required: true
         type: string
+      open_pr:
+        description: 'Whether to a PR'
+        required: false
+        type: boolean
+        default: true
       draft:
         description: 'Open a Draft PR'
         required: false
         type: boolean
         default: false
+      branch:
+        description: 'Git branch to use (defaults to "gobot/go-update-<go_version>")'
+        required: false
+        type: string
+        default: ''
+
+env:
+  BRANCH_NAME: ${{ inputs.branch || format('gobot/go-update-{0})', inputs.go_version) }}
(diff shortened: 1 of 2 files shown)


GustavoCaso:
For reference, this is an example PR created using this workflow https://github.com/DataDog/datadog-agent-buildimages/pull/499

GustavoCaso (approved):
LGTM 🔥

.github/workflows/go-update.yml
```diff
@@ -0,0 +1,79 @@
...
+
+on:
+  workflow_dispatch:
+    inputs:
+      go_version:
+        description: 'Go version'
+        required: true
+        type: string
+      open_pr:
+        description: 'Whether to a PR'
```
chouetz, line 11:
```suggestion
        description: 'Whether to open a PR'
```

.github/workflows/go-update.yml
```diff
@@ -0,0 +1,79 @@
...
+      branch:
+        description: 'Git branch to use (defaults to "gobot/go-update-<go_version>")'
+        required: false
+        type: string
+        default: ''
+
+env:
+  BRANCH_NAME: ${{ inputs.branch || format('gobot/go-update-{0})', inputs.go_version) }}
+
+jobs:
```
chouetz, line 29: should we add a `if` condition here to restrict the permissions to trigger this workflow (for instance to a list of hardcoded `github.actor`, not sure we can dynamically target a team)

chouetz (approved):

.github/workflows/go-update.yml
pgimalac: Nice, thanks !

.github/workflows/go-update.yml
pgimalac: I believe workflows can only be triggered by someone having `write` access to the repo, so if someone who's not supposed to can trigger it we have a big issue already :smile:

pgimalac: I checked and there is an action to check team membership, the question now is do we want to ?
I don't think we need to as workflows need write permission to be triggered anyway

chouetz: I agree it's an additional verification. However, you maybe don't want anyone having write permission to trigger it :)

pgimalac: > However, you maybe don't want anyone having write permission to trigger it :)

That's a good remark, but considering the created PR would need reviews anyway, and you can do the exact same changes manually if you'd like, I would say it's not needed.

PR #501, opened 2023-12-26
[ASCII-999] Single source of truth for Go version

pgimalac:
### What does this PR do?

Move all definitions of Go versions and Go archive SHA256 to a single file.

### Motivation

Simplify Go updates (and considerably simplify the `update-go` invoke task), and make it possible to get the current Go version (as there is only one).

The version of Go used in the datadog-agent repository also relies on a single file, so I think this makes sense.


dustmop (approved):
LGTM, cool!

build-container.ps1
```diff
@@ -32,6 +32,13 @@ if($Tag -eq $null -or $Tag -eq ""){
 }
 $arglist += "build"
 
+# Read arguments from go.env file
+$lines = Get-Content -Path 'go.env'
+foreach ($line in $lines) {
+    $arglist += "--build-arg"
+    $arglist += "$line"
```
KevinFairise2, line 39: Did you test it work as expected? I feel like we will lack of some spaces in the constructed string?

tasks/update_go.py
```diff
@@ -11,53 +11,14 @@
 
 # list all platforms used in the repo
 # store the platform names in the format expected by Go
-# the variables in the Dockerfiles should be stored in uppercased variables
 # see https://go.dev/dl/ for the full list of platforms
-PLATFORMS: Set[Platform] = {
+PLATFORMS: List[Platform] = {
```
KevinFairise2, line 15: Is it working to type it as a `List` and define it with `{ }`. Should we change it to `[ ]` ?

windows/Dockerfile
```diff
@@ -27,6 +27,7 @@ ARG GIT_VERSION
 ARG SEVENZIP_VERSION
 ARG VS2017BUILDTOOLS_VERSION
 ARG GO_VERSION
+ARG GO_SHA256_WINDOWS_AMD64
```
KevinFairise2, line 30: According to what we do on other Dockerfile I think you would need to add something like `ENV GO_SHA256_WINDOWS_AMD64=$GO_SHA256_WINDOWS_AMD64` to have it as an environment variable that you can use in the ps scripts that are called in the Dockerfile.

I am not sure this is absolutely required tho, if it works like that feel free to ignore the comment

KevinFairise2 (commented):
Some comments but that's a great improvement thanks!

build-container.ps1
pgimalac: I was also confused at first, but the thing is that `arglist` is a list, so I guess `+=` adds a new element at the end, it's not string concatenation

Note that this logic is basically copy pasted this already existing piece of code https://github.com/DataDog/datadog-agent-buildimages/blob/a4236ffb5b418a69dcc3c8c2f5c63c6a76397144/build-container.ps1#L37-L38

pgimalac: I didn't test it manually but the pipeline is green, including the [pipeline](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/pipelines/25601992) triggered in `datadog-agent`

gjulianm (approved):

build-container.ps1
KevinFairise2: Ah yes you're right, this is not string concatenation! Looks good then!

windows/Dockerfile
pgimalac: The main difference between ENV and ARG is that ARG is build time only while ENV is still accessible at run time.
Both declare environment variables, what change is just when it is accessible, in particular they are both accessible when running a script during the build of the `Dockerfile`.

The following is just a workaround so that an argument persist at run time:
```
ARG name=default
ENV name $name
```

We don't need `GO_SHA256_WINDOWS_AMD64` to be accessible at run time, so no need for an `ENV`.

See https://vsupalov.com/docker-arg-env-variable-guide for a very interesting description of the difference.

tasks/update_go.py
pgimalac: It was disabled :smile:

KevinFairise2 (approved):

windows/Dockerfile
KevinFairise2: Oh that's good to know, we could probably clean some of the Dockerfile then. Thanks for the explanation and the link!

pgimalac: Yes I thought so too :)

.gitlab-ci.yml
```diff
@@ -108,7 +108,8 @@ get_agent_version:
     # ECR login
     - eval "$(aws ecr get-login --region us-east-1 --no-include-email --registry-ids 486234852809)"
     # Build
-    - docker build --build-arg BASE_IMAGE=$BASE_IMAGE --build-arg DD_TARGET_ARCH=$DD_TARGET_ARCH $CUSTOM_BUILD_ARGS --tag 486234852809.dkr.ecr.us-east-1.amazonaws.com/ci/datadog-agent-buildimages/$IMAGE${ECR_TEST_ONLY}:v$CI_PIPELINE_ID-${CI_COMMIT_SHORT_SHA} --file $DOCKERFILE .
+    - GO_BUILD_ARGS=$(cat go.env | sed -e 's/^/--build-arg /' | tr '\n' ' ')
```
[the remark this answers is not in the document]
pgimalac: The `build` job is for Unix, there is a `winbuild` equivalent for `Windows` jobs

clarkb7 (changes requested):
`install-all.ps1` won't work on its own anymore with the `GO_` options removed from `versions.ps1`  

Can you add something similar to what you added to `build-container.ps1` to either `install-all.ps1` or to `versions.ps1` ?

It looks like this loop from `install-all.ps1` is similar to what is in `build-container.ps1`.
https://github.com/DataDog/datadog-agent-buildimages/blob/a4236ffb5b418a69dcc3c8c2f5c63c6a76397144/windows/install-all.ps1#L24-L28

pgimalac:
@clarkb7 you are right I didn't consider `install-all.ps1` could be run by itself, I fixed it (still waiting for the CI)

PR #506, opened 2024-01-03
Upgrade the version of Python to 3.11

FlorentClarret:
- Switch from miniconda to miniforge on x64, similar to what we already do on aarch64. It simplifies the script quite a bit
- Bump python version from py3.9 to 3.11
- Manually install openssl on `armhf` architectures because Python 3.11 now requires it

Note: I had a ton of compilation errors with miniconda on centos6, using miniforge seems to work just fine

We will need to merge https://github.com/DataDog/datadog-agent/pull/21844 as well

Green CI with [this commit](https://github.com/DataDog/datadog-agent-buildimages/pull/506/commits/949dbc2f25dc1d981a475cbfa4974be92ac0aac3): https://gitlab.ddbuild.io/DataDog/datadog-agent-buildimages/-/pipelines/25947899. I don't know why the child pipeline picked up main, which made fail the windows job because of the path to the awscli binary

https://datadoghq.atlassian.net/browse/AITS-324


release.json
```diff
@@ -1,3 +1,3 @@
 {
-    "agent_version": "main"
+    "agent_version": "florentclarret/python_3_11_build_image"
```
FlorentClarret, line 2:
```suggestion
    "agent_version": "main"
```

setup_python.sh
```diff
@@ -38,20 +32,32 @@ case $DD_TARGET_ARCH in
...
+    pushd /Python-$PY3_VERSION
+        ./configure --with-openssl=/opt/openssl --with-openssl-rpath=auto
+        make -j $(nproc)
         make install
     popd
-    rm -rf Python-3.9.5
-    rm Python-3.9.5.tgz
-    ln -sf /usr/bin/python3.9 /usr/bin/python3
+    rm -rf Python-$PY3_VERSION Python-$PY3_VERSION.tgz
+    ln -sf /usr/bin/python3.11 /usr/bin/python3
```
chouquette, line 60: Is this meant to be the python executable you just built? Or another version installed some other way?

FlorentClarret: Yes it is the Python executable we just built

chouquette: Then I think something is wrong, by default (ie. without `--prefix`) python should be installed in `/usr/local/`

FlorentClarret: I'll pull the image and double check then! But if that's wrong, it's already broken 😅

FlorentClarret: you're right, we are using /usr/local/bin/python3 and `/usr/bin/python3` is a broken link for the exact reason you mention. I'll drop the `ln`

chouquette: Thanks for checking! I was a bit confused as to why it worked if the executable wasn't there, but if it uses it from /usr/local directly then it makes sense :)

clarkb7 (approved):

amenasria (approved):
**👏 praise:** ‏Thanks for this ! Letsgoo !

PR #507, opened 2024-01-04
Change always to on_success to prevent running child pipelines when images are not built

KevinFairise2:
With `always` the jobs are always run even if previous stages failed. On success will run jobs only if previous stages succeeded. It should avoid situations where we run datadog-agent pipeline with images that failed their build.
It will make the behaviour closer to what we have before https://github.com/DataDog/datadog-agent-buildimages/pull/497

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -43,7 +43,7 @@ variables:
     when: never
   - if: $CI_COMMIT_TAG != null
     when: never
-  - when: always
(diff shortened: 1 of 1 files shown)


Pythyu (approved):

Pythyu:
It might break `wait_for_tests` when failling actually, a separate reference for `wait_for_tests` might be needed with `always` still on

PR #510, opened 2024-01-08
don't expose build only variables to the containers

chouquette:
We don't need our dependencies versions and their tarballs checksum to be available when running these containers


KevinFairise2 (approved):

chouquette:
as it turns out, `GCC_VERSION` is used by our build scripts to compose the path to the toolchain, so I readded it to the environment

PR #511, opened 2024-01-09
rpm-armhf: update git

chouquette:
We need a version that supports `-C` for omnibus caching, and this will make our build images more consistant with each others

### DIFF
--- a/rpm-armhf/Dockerfile
+++ b/rpm-armhf/Dockerfile
@@ -14,6 +14,8 @@ ARG GO_VERSION
 ARG GO_SHA256_LINUX_ARMV6L
 ARG DD_TARGET_ARCH=armhf
 ARG BUNDLER_VERSION=2.4.20
+ARG GIT_VERSION=2.10.1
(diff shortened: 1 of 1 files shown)


Pythyu (approved):
Looks good ! Just a small question for context how fetching this specific git version matter for the `-C` caching option of omnibus ?

Pythyu:
> > Looks good ! Just a small question for context how fetching this specific git version matter for the `-C` caching option of omnibus ?
> 
> Honest answer: I don't know which version is currently installed and which version introduced the `-C` option, however it currently isn't recognized: https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/402922445 Since the armhf job is the only one with the issue, I just copied the existing install from another image, but we might want to check if there are other improvements we could benefit from since 2.10 is rather old (2016)

I see make sense ! Thanks for the context 😄

chouquette:
Test job with the test image: https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/403750776

PR #512, opened 2024-01-10
[CI] Better image Slack notifications

amenasria:
This PR enhances the image Slack notifications

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -515,9 +515,9 @@ notify-images-available:
     - when: on_success
   dependencies: []
   script: |
-    export MESSAGE="Your :docker: images with tag `v$CI_PIPELINE_ID-${CI_COMMIT_SHORT_SHA}` are ready.
+    export MESSAGE="Your :docker: images with tag \`v$CI_PIPELINE_ID-${CI_COMMIT_SHORT_SHA}\` are ready.
     :idea: You can test them in the datadog-agent repository by running:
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -515,7 +515,10 @@ notify-images-available:
     - when: on_success
   dependencies: []
   script: |
-    export MESSAGE="Your images :docker: tagged 'v$CI_PIPELINE_ID-${CI_COMMIT_SHORT_SHA}' are ready. :idea: In datadog-agent you can test with 'inv pipeline.update-buildimages -i v$CI_PIPELINE_ID-${CI_COMMIT_SHORT_SHA} [--no-test-version] [--branch-name <your_branch>]'"
+    export MESSAGE="Your :docker: images with tag \`v$CI_PIPELINE_ID-${CI_COMMIT_SHORT_SHA}\` are ready.
+    :idea: You can test them in the datadog-agent repository by running:
+    \`\`\`inv pipeline.update-buildimages -i v$CI_PIPELINE_ID-${CI_COMMIT_SHORT_SHA} [--no-test-version] [--branch-name <your_branch>]\`\`\`
```
chouquette, line 520: Should we remove the square brackets around `--no-test-version`?

This notification is only sent when running on a pipeline branch, meaning the command line to test the resulting image will always need to contain `--no-test-version`

Removing the `[ ]` makes copy pasting the command easier for lazy people such as myself :innocent:

chouquette (approved):

.gitlab-ci.yml
amenasria: Discussed offline: `--no-test-version` is doing the opposite i.e removing the `_test_only` suffix from the image, so we don't want that. Marking this as solved !

PR #521, opened 2024-01-26
fix(all): $HOME is not known in docker context, force it to root home

chouetz:
This will enable the auto-setup for curl/wget/aws and should increase the datadog-agent ci resilience.
Do not necessarily remove the redundant options from curl commands in scripts and dockerfiles

### DIFF
--- a/.curlrc
+++ b/.curlrc
@@ -1,6 +1,6 @@
 --continue-at -
---include
 --location
 --retry 10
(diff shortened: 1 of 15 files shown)


chouquette (approved):
LGTM!

gjulianm (approved):
LGTM

PR #550, opened 2024-03-15
rpm-x64: install perl from the distribution

chouquette:
it's now recent enough for our needs, and we don't need to build perl anymore


chouquette:
FWIW, this divides the RPM image build duration by 2, and drops around 300MB from the image

spencergilbert (approved):

PR #552, opened 2024-03-22
[CI] Install `codecov` cli through pip

amenasria:
With https://github.com/DataDog/datadog-agent/pull/23680 we're installing codecov cli through `pip`, we don't need the binary install in the image anymore.

### DIFF
--- a/circleci/Dockerfile
+++ b/circleci/Dockerfile
@@ -62,18 +62,6 @@ RUN set -ex \
     && mv $CMAKE_NAME $CMAKE_DEST_DIR \
     && rm $CMAKE_ARCHIVE
 
-# Codecov
-ENV CODECOV_VERSION=0.6.1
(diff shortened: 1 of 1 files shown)


CelianR (approved):

KevinFairise2 (approved):
I do not have a strong opinion on what the best thing is. But with that we go from `codecov` being installed once in the image, to `codecov` being installed on every job that needs.

amenasria:
It's still installed in the image just in [a different place](https://github.com/DataDog/datadog-agent-buildimages/pull/552/files#diff-f7e8f6b6f3b8fcbbd916bfa689d3405520116af3abbf1a790013fb4f8d2abcfeR70) !

amenasria:
Oh you're actually right !

amenasria:
Change of plan, I'll use the bash install for all images, instead of codecov wheel.

PR #553, opened 2024-03-29
[gitlab-use-module] Added requirements

CelianR:
Added python-gitlab for this PR : https://github.com/DataDog/datadog-agent/pull/24070

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,8 +11,9 @@ docker-squash
 dulwich
 invoke
 packaging
+python-gitlab
 reno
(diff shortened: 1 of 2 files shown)


chouquette:
LGTM, let's wait on the CI to ensure the gitlab package doesn't pull incompatible dependencies (this was the case for github which is why we have to install it at runtime in each job :cry:)

KevinFairise2 (approved):

chouetz (approved):

PR #554, opened 2024-04-03
Use an alternate way of fetching xz-utils sources

chouquette:
This version isn't affected by the backdoor AFAWK, and the backdoor was only present in the source tarballs provided by xz-utils, generated through `make dist`, as far as I understood.

### DIFF
--- a/deb-x64/Dockerfile
+++ b/deb-x64/Dockerfile
@@ -92,6 +92,21 @@ RUN curl -LO https://ftp.gnu.org/gnu/gettext/gettext-${GETTEXT_VERSION}.tar.xz \
     && cd / \
     && rm -rf "gettext-${GETTEXT_VERSION}"
 
+# Git
+RUN curl -OL https://www.kernel.org/pub/software/scm/git/git-${GIT_VERSION}.tar.gz \
+    && echo "${GIT_SHA256}  git-${GIT_VERSION}.tar.gz" | sha256sum --check \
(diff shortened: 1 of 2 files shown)


deb-x64/Dockerfile
```diff
@@ -94,14 +93,18 @@ RUN curl -LO https://ftp.gnu.org/gnu/gettext/gettext-${GETTEXT_VERSION}.tar.xz \
...
 
 # Now install a recent enough liblzma (5.2+) which supports parallel compression
-RUN curl -LO https://github.com/tukaani-project/xz/releases/download/v${LIBLZMA_VERSION}/xz-${LIBLZMA_VERSION}.tar.xz \
-    && echo "${LIBLZMA_SHA256} /xz-5.2.11.tar.xz" | sha256sum --check \
-    && tar -xf /xz-${LIBLZMA_VERSION}.tar.xz \
-    && cd xz-${LIBLZMA_VERSION} \
+RUN if [ "$DD_TARGET_ARCH" = "armhf" ] ; then \
+    LIB_DEST="arm-linux-gnueabihf"; \
+    else \
+    LIB_DEST="aarch64-linux-gnu"; \
```
KevinFairise2, line 99: Why do we now have reference to `aarch64` in the `x64` build image?

KevinFairise2: Hmm if I understand correctly. I think you just copy pasted what we had in the `arm64` dockerfile, right? Probably need some changes

chouquette: Because I wrongfully copy pasted the same block as I forgot there was an arm specificity :facepalm:

PR #558, opened 2024-04-04
Update the Rust toolchain to `1.76.0` on all Linux builders.

remeh:
In order to build the SDS shared library in the datadog-agent pipeline, we need a more recent Rust.

Rust was used to compile `python-cryptography` used by integrations in the Agent, but it won't be the case for next releases (and it was compatible with recent Rust anyway).


deb-arm/Dockerfile
```diff
@@ -112,6 +120,21 @@ RUN if [ "$DD_TARGET_ARCH" = "armhf" ] ; then \
 ENV PATH="/usr/local/go/bin:${PATH}"
 ENV PATH="${GOPATH}/bin:${PATH}"
 
+# Rust is needed to compile the SDS library
+RUN if [ "$DD_TARGET_ARCH" = "aarch64" ] ; then curl -sSL -o rustup-init https://static.rust-lang.org/rustup/archive/${RUSTUP_VERSION}/${RUSTUP_ARCH_AARCH64}/rustup-init \
+    && echo "${RUSTUP_SHA256_AARCH64}  rustup-init" | sha256sum --check \
+    && chmod +x ./rustup-init \
+    && ./rustup-init -y --profile minimal --default-toolchain ${RUST_VERSION} \
+    && echo "${RUSTC_SHA256_AARCH64}  $HOME/.cargo/bin/rustc" | sha256sum --check \
+    && rm ./rustup-init; fi
```
chouquette, line 129: Just double checking, does this leave some kind of temporary files behind? We had a few GB wasted by go temporary files in other build images, I'd rather not have to download those for rust :sweat_smile:

deb-arm/Dockerfile
```diff
@@ -112,6 +120,21 @@ RUN if [ "$DD_TARGET_ARCH" = "armhf" ] ; then \
...
+RUN if [ "$DD_TARGET_ARCH" = "armhf" ] ; then curl -sSL -o rustup-init https://static.rust-lang.org/rustup/archive/${RUSTUP_VERSION}/${RUSTUP_ARCH_ARMHF}/rustup-init \
+    && echo "${RUSTUP_SHA256_ARMHF}  rustup-init" | sha256sum --check \
+    && chmod +x ./rustup-init \
+    && ./rustup-init -y --profile minimal --default-toolchain ${RUST_VERSION} \
+    && echo "${RUSTC_SHA256_ARMHF}  $HOME/.cargo/bin/rustc" | sha256sum --check \
+    && rm ./rustup-init; fi
```
chouquette, line 135: I think this would be more concise if you were conditionally assigning the triplet/shasum to a common couple of variables and only had a single install instruction

deb-x64/Dockerfile
```diff
@@ -19,7 +19,7 @@ ARG CMAKE_SHA256="5eaf3ded8255b3d149c7aad68c473f01c2345dce872e03c92f2e9c0185dfa7
 ARG CLANG_VERSION=8.0.0
 ARG CLANG_SHA256="9ef854b71949f825362a119bf2597f744836cb571131ae6b721cd102ffea8cd0"
 ARG DD_TARGET_ARCH=x64
-ARG RUST_VERSION=1.64.0
+ARG RUST_VERSION=1.76.0
 ARG RUSTC_SHA256="0b2f6c8f85a3d02fde2efc0ced4657869d73fccfce59defb4e8d29233116e6db"
```
chouquette, line 23: Doesn't this imply a change of checksum?

chouquette (commented):
> but it won't be the case for next releases
Nice trailer, now I want to know more :eyes: 

LGTM, I'm just a bit concerned by the size increase, it seems we're going from `1.48 GB` to `2.33` for the `deb-armhf` image, does that seem like the expected ballpark to you?

deb-arm/Dockerfile
remeh: I'll do this.

deb-arm/Dockerfile
chouquette: Nah as long as we only ship the final binaries and not all the intermediary objects I think we're good :)

deb-x64/Dockerfile
chouquette: This is super confusing to me, but if it works it works :sweat_smile:

deb-arm/Dockerfile
remeh: For a future reader: Correction 14MB per binary and as they're hard-links, they only count once for the inode size.
The toolchain is 800MB (most of it are the `.so` _used_ by the Rust toolchain), there won't be much I can do.

deb-arm/Dockerfile
```diff
@@ -112,6 +123,23 @@ RUN if [ "$DD_TARGET_ARCH" = "armhf" ] ; then \
 ENV PATH="/usr/local/go/bin:${PATH}"
 ENV PATH="${GOPATH}/bin:${PATH}"
 
+# Rust is needed to compile the SDS library
+RUN if [ "$DD_TARGET_ARCH" = "armhf" ] ; then \
+           export RUSTUP_SHA256=${RUSTUP_SHA256_ARMHF} \
+        && export RUSTUP_ARCH=${RUSTUP_ARCH_ARMHF} \
+        && export RUSTC_SHA256=${RUSTC_SHA256_ARMHF} \
```
chouquette, line 130: small nit: you could simplify by doing `export FOO=bar BAR=baz ...` in a single command, which should reduce the boilerplate a bit

chouquette (approved):

deb-arm/Dockerfile
```diff
@@ -112,6 +123,23 @@ RUN if [ "$DD_TARGET_ARCH" = "armhf" ] ; then \
 ENV PATH="/usr/local/go/bin:${PATH}"
 ENV PATH="${GOPATH}/bin:${PATH}"
 
+# Rust is needed to compile the SDS library
```
pducolin, line 126: **❓ question**
What is SDS?

pducolin (approved):

deb-arm/Dockerfile
remeh: Here's a link to the [Sensitive Data Scanner](https://docs.datadoghq.com/sensitive_data_scanner/?tab=addscanningrulefromthelibraryrules) doc.

PR #559, opened 2024-04-05
Add ddpy3/bin to $PATH in sysprobe images

gjulianm:
With #547 it seems that some binaries that our code (CI but also local tests) expected to be in `$PATH` cannot be found anymore. This PR adds the binary folder for the ddpy3 virtualenv to `$PATH` to avoid that problem.

Agent pipeline for validation: https://gitlab.ddbuild.io/DataDog/datadog-agent/-/pipelines/31522447


CelianR (approved):

usamasaqib (approved):
The KMT tests in the linked pipeline did not run.

gjulianm:
> The KMT tests in the linked pipeline did not run.

I ran them manually and forgot system-probe ones, they seem to run ok now, thanks for the catch!

PR #565, opened 2024-04-09
feat(ci): add uploader to windows image

chouetz:
ACIX-192


windows/helpers/phase3/install_ci_uploader.ps1
```diff
@@ -0,0 +1,22 @@
...
+$isInstalled, $isCurrent = Get-InstallUpgradeStatus -Component "datadog-ci" -Keyname "version" -TargetValue $Version
+if($isInstalled -and $isCurrent) {
+    Write-Host -ForegroundColor Green "Datadog-ci uploader up to date"
+    return
+}
+if($isInstalled -and -not $isCurrent){
+    Remove-Item -Recurse -Force c:\datadog-ci -ErrorAction SilentlyContinue
+}
+$source="https://github.com/DataDog/datadog-ci/releases/download/v${Version}/datadog-ci_win-x64"
+$target = "c:\datadog-ci"
```
CelianR, line 17: I'm not used to Windows but isn't it in uppercase usually ?
```suggestion
$target = "C:\datadog-ci"
```

CelianR (commented):
I'm not sure I have enough experience in this to approve it but lgtm

clarkb7 (approved):
approving with a note

Would like to see it installed somewhere besides `C:\`, but it looks like a bunch of other tools also install to `C:\` 😞 so it's fine.

If you're up for it, it looks like we put our [signing tool](https://github.com/DataDog/datadog-agent-buildimages/blob/ade68c7e0878c9013d1cce6387f183da645dee36/windows/helpers/phase3/install_winsign.ps1) in `C:\devtools`.

windows/helpers/phase3/install_ci_uploader.ps1
clarkb7: usually by convention yes, but paths are technically case insensitive

julien-lebot: Windows should be case insensitive

PR #566, opened 2024-04-10
Build rpmbuild with parallel lzma support

chouquette:
This PR replicates what's done for other distributions by enabling parallel compression for XZ.

This is expected to save around 10 minutes per SUSE job in datadog-agent pipelines, as they currently take ≃15 minutes instead of ≃5 for other distributions: 
image

### DIFF
--- a/suse-x64/Dockerfile
+++ b/suse-x64/Dockerfile
@@ -175,7 +175,7 @@ RUN curl -LO https://ftp.gnu.org/gnu/gettext/gettext-${GETTEXT_VERSION}.tar.xz \
     && cd "gettext-${GETTEXT_VERSION}" \
     && ./configure --prefix=/usr && make -j$(nproc) && make install \
     && cd / \
-    && rm -rf "gettext-${GETTEXT_VERSION}"
+    && rm -rf "gettext-${GETTEXT_VERSION}*"
 
 # Now install a recent enough liblzma (5.2+) which supports parallel compression
 RUN git clone -b v${LIBLZMA_VERSION} https://github.com/tukaani-project/xz.git \
@@ -194,7 +194,7 @@ RUN curl -sL -o /tmp/rpm-4.14.3.tar.bz2 http://ftp.rpm.org/releases/rpm-4.14.x/r
     && make -j$(nproc) \
     && make install \
     && ldconfig \
-    && rm -rf /tmp/rpm-4.14.3
+    && rm -rf /tmp/rpm-4.14.3*
 
 # Rebuild RPM database with the new rpm
 RUN mkdir -p /usr/local/var/lib/rpm \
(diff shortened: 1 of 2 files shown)


suse-x64/Dockerfile
```diff
@@ -165,25 +168,39 @@ RUN curl -sL -O "https://ftp.gnu.org/gnu/binutils/binutils-${BINUTILS_VERSION}.t
...
 ENV PATH="/usr/local/binutils/bin:$PATH"
 
+# xz needs gettext 0.19.6 or newer
+RUN curl -LO https://ftp.gnu.org/gnu/gettext/gettext-${GETTEXT_VERSION}.tar.xz \
+    && echo "${GETTEXT_SHA256}  gettext-${GETTEXT_VERSION}.tar.xz" | sha256sum --check \
+    && tar xf "gettext-${GETTEXT_VERSION}.tar.xz" \
+    && cd "gettext-${GETTEXT_VERSION}" \
+    && ./configure --prefix=/usr && make -j$(nproc) && make install \
+    && cd / \
+    && rm -rf "gettext-${GETTEXT_VERSION}"
```
FlorentClarret, line 178: nit - I think you do not remove the `tar.gz` file, just the folder.

suse-x64/Dockerfile
```diff
@@ -165,25 +168,39 @@ RUN curl -sL -O "https://ftp.gnu.org/gnu/binutils/binutils-${BINUTILS_VERSION}.t
...
     && echo "13711268181a6e201eb9d4eb06384a7fcc42dcc6c9057a62c53472cfc5ba44b5  rpm-4.14.3.tar.bz2" | sha256sum --check \
     && tar -xjf /tmp/rpm-4.14.3.tar.bz2 \
     && cd rpm-4.14.3/ \
     && ./configure --with-posixmutexes --with-external-db --without-lua CPPFLAGS="-I/usr/include/nspr4 -I/usr/include/nss3 -I/usr/include/db4"\
-    && make \
+    && make -j$(nproc) \
     && make install \
-    && ldconfig
+    && ldconfig \
+    && rm -rf /tmp/rpm-4.14.3
```
FlorentClarret, line 197: Same here with the tar file?

FlorentClarret (approved):
LGTM, just left one question/comment

suse-x64/Dockerfile
chouquette: Good point

suse-x64/Dockerfile
chouquette: Also a good point

suse-x64/Dockerfile
chouquette: Done, and replicated to the debian dockerfile which I copy pasted from

chouquette:
image
The results are as expected (the 2 top lines are packaging durations for SUSE x86_64 jobs)

PR #568, opened 2024-04-11
Add ruff to the Python deps

FlorentClarret:
We plan to replace flake8 + isort + black with [ruff](https://pypi.org/project/ruff/).

I'll open another PR in the agent later to use it, and another one after to drop the old libraries


CelianR (approved):
Are you sure you want to only update constraints and not requirement files ? If so lgtm 👍

FlorentClarret:
> Are you sure you want to only update constraints and not requirement files ? If so lgtm 👍

Fixed!

PR #576, opened 2024-04-16
Add gitlab_agent_deploy image

CelianR:
This image was from the builders repository, it is moved in build images.

In addition:
1. The Python version has been updated from 3.7 to 3.8.
2. The requirements of buildimages are local requirements, not those from the main branch

Note: The new job will be set as required after merge


agent-deploy/requirements.txt
chouetz, line 1: Should we put this in the `requirements` folder and put constraints with the other files?

CelianR: Yes good idea

chouetz (approved):

PR #580, opened 2024-04-24
[CI] Install Codecov on [`deb`,`rpm`] x [`amd`,`arm`] images

amenasria:
We want to start running codecov on every OS, so I'm adding codecov on all images running the go tests.


setup_codecov.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+set -ex
+
+CODECOV_VERSION=0.6.1
+
+# Integrity checking the uploader
+curl https://keybase.io/codecovsecurity/pgp_keys.asc | gpg --no-default-keyring --keyring trustedkeys.gpg --import
```
spencergilbert, line 8: TIL keybase still exists.

spencergilbert (approved):

FlorentClarret (approved):
LGTM!

setup_codecov.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+set -ex
```
chouetz, line 3: the `x` is risky, isn't it? Why not the usual 
```suggestion
set -euo pipefail
```

chouetz (approved):

setup_codecov.sh
amenasria: Good point !

PR #581, opened 2024-05-02
Remove IBM MQ build images

alopezz:
Starting with https://github.com/DataDog/datadog-agent/pull/23094, we build Python dependencies without using these build images, so IBM MQ is no longer needed, and Rust is no longer associated to building Python libraries either (updated related comments).

See also https://github.com/DataDog/datadog-agent-macos-build/pull/187


FlorentClarret (approved):
YAY!

PR #600, opened 2024-06-03
rpm-arm64: upgrade to a newer `binutils` version.

remeh:
In order to remove:

```
unsupported GNU_PROPERTY_TYPE (5) type: 0xc0000000
```

from `ld` calls, see these errors: 

It is similar to what's done in [rpm-x64](https://github.com/DataDog/datadog-agent-buildimages/blob/2333b0b25faf155b31e4e054aad5677c6ab62008/rpm-x64/Dockerfile#L131-L142) and [suse-x64](https://github.com/DataDog/datadog-agent-buildimages/blob/2333b0b25faf155b31e4e054aad5677c6ab62008/suse-x64/Dockerfile).

### DIFF
--- a/rpm-arm64/Dockerfile
+++ b/rpm-arm64/Dockerfile
@@ -17,6 +17,8 @@ ARG RUSTC_SHA256="673e336c81c65e6b16dcdede33f4cc9ed0f08bde1dbe7a935f113605292dc8
 ARG RUSTUP_VERSION=1.26.0
 ARG RUSTUP_SHA256="673e336c81c65e6b16dcdede33f4cc9ed0f08bde1dbe7a935f113605292dc800"
 ARG RUSTUP_ARCH="aarch64-unknown-linux-gnu"
+ARG BINUTILS_VERSION="2.39"
+ARG BINUTILS_SHA256="d12ea6f239f1ffe3533ea11ad6e224ffcb89eb5d01bbea589e9158780fa11f10"
 ARG CMAKE_VERSION=3.23.0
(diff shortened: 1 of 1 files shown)


rpm-arm64/Dockerfile
```diff
@@ -103,6 +115,9 @@ RUN /bin/bash -l -c "rvm install 2.7 --with-openssl-dir=${CONDA_PATH} && rvm cle
 RUN /bin/bash -l -c "gem install bundler --version $BUNDLER_VERSION --no-document"
 RUN echo 'source /usr/local/rvm/scripts/rvm' >> /root/.bashrc
 
+# Override updated linker system-wide
+RUN ln -sf /usr/local/binutils/bin/ld /usr/bin/ld
```
chouquette, line 119: Won't this create problem if we end up using another binutils tool from the distribution?

remeh: I'll be honest: I don't know yet. It is what's done in [rpm-x64](https://github.com/DataDog/datadog-agent-buildimages/blob/2333b0b25faf155b31e4e054aad5677c6ab62008/rpm-x64/Dockerfile#L131-L142) and [suse-x64](https://github.com/DataDog/datadog-agent-buildimages/blob/2333b0b25faf155b31e4e054aad5677c6ab62008/suse-x64/Dockerfile).

I'll see if I can use it only on the problematic part instead of this `ln -sf`

chouquette: Oh if we've been doing so for at least a year I suppose it's fine enough for the time being

chouquette (approved):

PR #604, opened 2024-06-07
Get the go version from the `go.env` file to build the devcontainer

FlorentClarret:
So we don't need to manually bump it anymore

Follows https://github.com/DataDog/datadog-agent-buildimages/pull/603


.gitlab/devcontainer.yml
```diff
@@ -16,7 +16,7 @@ lint_devcontainer:
     matrix:
       - PLATFORM: [ amd64,arm64 ]
   script:
-    - docker buildx build --no-cache --platform linux/${PLATFORM} --tag ${DOCKER_TARGET}-${PLATFORM} -f devcontainer/Dockerfile . ${PUSH}
+    - docker buildx build --build-arg GO_VERSION=$(grep GO_VERSION go.env | cut -d = -f 2) --no-cache --platform linux/${PLATFORM} --tag ${DOCKER_TARGET}-${PLATFORM} -f devcontainer/Dockerfile . ${PUSH}
```
chouetz, line 19: Could/should we do exactly [like the others](https://github.com/DataDog/datadog-agent-buildimages/blob/main/.gitlab-ci.yml#L116-L117)?

chouetz (approved):

.gitlab/devcontainer.yml
FlorentClarret: I switched to this way but kept the grep because I don't need the other variables! Thanks

PR #607, opened 2024-06-12
Use mirrors for MSYS2

julien-lebot:
The main MSYS2 repo seems down. I've added 2 synthetics monitor as well to monitor those endpoints.

https://datadoghq.atlassian.net/browse/WINA-861


amenasria (approved):
LGTM !

PR #609, opened 2024-06-14
upgrade system-probe images to bullseye for gcc 9

brycekahle:
Go 1.22 needs gcc 9 to build the race detector and gcc 9 is not available in `buster`, so we are upgrading to `bullseye`. This has been tested in the agent repo and did not cause any test failures.


usamasaqib (approved):

amenasria (approved):
LGTM as long as the CI passes !

PR #612, opened 2024-06-19
Bump datadog-ci and include it in ARM images

gjulianm:
As of version 2.38, we have ARM64 builds of `datadog-ci`. This PR adds `datadog-ci` to ARM64 images if the corresponding x64 image had it too, and also bumps the version to the latest in all images.

Validated in pipeline https://gitlab.ddbuild.io/DataDog/datadog-agent/-/pipelines/37109034


vboulineau (approved):

deb-arm/Dockerfile
```diff
@@ -74,6 +76,12 @@ RUN if [ "$DD_TARGET_ARCH" = "aarch64" ] ; then \
     fi
 RUN chmod +x /usr/local/bin/curl
 
+# CI uploader, to send junit to ci-visibility during CI tests
+RUN curl -fsSL https://github.com/DataDog/datadog-ci/releases/download/v${CI_UPLOADER_VERSION}/datadog-ci_linux-arm64 --output "/usr/local/bin/datadog-ci" && \
```
chouquette, line 80: I think you need to ensure that the arch is indeed arm64 as this Dockerfile is also used for the `deb_armhf` image

gjulianm: Fixed, thanks for the warning!

usamasaqib (approved):

deb-arm/Dockerfile
chouquette: My pleasure, thanks for fixing :)

chouquette (approved):

PR #615, opened 2024-06-24
[Codecov] Install `codecov-cli` instead of the Codecov uploader

amenasria:
Why we're replacing the legacy uploader by the `codecov-cli` binary:

TL; DR: the 'uploader' has been deprecated since 2023 (see screen below). The right way to upload coverage files to Codecov now is to use the codecov-cli binary. I kept it before as it was the legacy way we were using Codecov, but we're now catching up with upstream because:

- We're missing interesting features:
    - Have the UI build URLs redirecting directly to our gitlab jobs.
    - Be able to get Codecov information right after you pushed to remote (this will be needed for later as it might be a mandatory check in the CI).
- Security
- Maintenance / ease of setup (doc is outdated sometimes)


setup_codecov.sh
```diff
@@ -2,16 +2,16 @@
 
 set -eo pipefail
 
-CODECOV_VERSION=0.6.1
+CODECOV_VERSION=0.7.1
 
 # Determine the Architecture to install the Codecov uploader
 if [ -z "$DD_TARGET_ARCH" ]; then
-  echo "DD_TARGET_ARCH environment variable is not set. The Codecov x64 uploader will be installed by default."
+  echo "DD_TARGET_ARCH environment variable is not set. The x64 version of Codecov CLI will be installed by default."
```
julien-lebot, line 9: Could you add a comment to clarify why 'CLI' is needed over 'uploader'?

julien-lebot (approved):

setup_codecov.sh
amenasria: I can answer here, but IMO it would be very confusing to have this info in the code.

TL; DR: the 'uploader' have been deprecated since 2023 (see screen below). The right way to upload coverage files to Codecov now is to use the `codecov-cli` binary. I kept it before as it was the legacy way we were using Codecov, but we're now catching up with upstream because:
- We're missing interesting features:
    - Have the UI build URLs redirecting directly to our gitlab jobs.
    - Be able to get Codecov information right after you pushed to remote (this will be needed for later as it might be a mandatory check in the CI).
- Security
- Maintenance / ease of setup (doc is outdated sometimes)
image

julien-lebot: Thanks for the clarifications

amenasria: I added this in the PR description, thanks for the reminder !🫰

PR #618, opened 2024-07-02
Switch from mirrorlist to vault repo on Centos 7

FlorentClarret:
Centos 7 EOLed on July 1st 2024 and mirrorlist.centos.org is no more. 

We need to switch to vault which is the historical archive

Note: the `btf_gen` job will be fixed in another PR

### DIFF
--- a/rpm-armhf/Dockerfile
+++ b/rpm-armhf/Dockerfile
@@ -39,6 +39,11 @@ RUN echo "armhfp" > /etc/yum/vars/basearch && \
 RUN rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
     rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-SIG-AltArch-Arm32
 
+# Centos 7 EOLed on July 1st, 2024. We need to switch to vault.centos.org to get the packages
+RUN sed -i s/mirror.centos.org/vault.centos.org/g /etc/yum.repos.d/*.repo && \
+    sed -i s/^#.*baseurl=http/baseurl=http/g /etc/yum.repos.d/*.repo && \
+    sed -i s/^mirrorlist=http/#mirrorlist=http/g /etc/yum.repos.d/*.repo
+
 # The last two lines contain dependencies for build of newer rpm
 RUN yum -y install @development which perl-ExtUtils-MakeMaker ncurses-compat-libs git procps \
     curl-devel expat-devel gettext-devel openssl-devel systemd-devel zlib-devel bzip2 glibc-static tar pkgconfig  \
--- a/rpm-armhf/Dockerfile
+++ b/rpm-armhf/Dockerfile
@@ -41,8 +41,8 @@ RUN rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
 
 # Centos 7 EOLed on July 1st, 2024. We need to switch to vault.centos.org to get the packages
 RUN sed -i s/mirror.centos.org/vault.centos.org/g /etc/yum.repos.d/*.repo && \
-    sed -i s/^#.*baseurl=http/baseurl=http/g /etc/yum.repos.d/*.repo && \
-    sed -i s/^mirrorlist=http/#mirrorlist=http/g /etc/yum.repos.d/*.repo
+    sed -i 's/^#.*baseurl=http/baseurl=http/g' /etc/yum.repos.d/*.repo && \
+    sed -i 's/^mirrorlist=http/#mirrorlist=http/g' /etc/yum.repos.d/*.repo
 
 # The last two lines contain dependencies for build of newer rpm
 RUN yum -y install @development which perl-ExtUtils-MakeMaker ncurses-compat-libs git procps \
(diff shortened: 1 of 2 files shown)


chouquette:
This looks OK but I was hoping to do the change in a common base image so we don't have to sed in a gazillion Dockerfile (https://github.com/DataDog/images/pull/5384)

chouquette (approved):
Approving so we can unbreak things, there will still be time to revisit/change the base image later on

FlorentClarret:
> This looks OK but I was hoping to do the change in a common base image so we don't have to sed in a gazillion Dockerfile ([DataDog/images#5384](https://github.com/DataDog/images/pull/5384))

I can try to switch to the image you have in this PR in a follow-up PR, see how it goes and drop the `sed` if it's working as expected. This way this repo will be fixed. WDYT?

KSerrania:
> This looks OK but I was hoping to do the change in a common base image so we don't have to sed in a gazillion Dockerfile (https://github.com/DataDog/images/pull/5384)

Note that if this common base image is in a private registry, then it is not suitable as a base for the images in this public repository.

rpm-armhf/Dockerfile
```diff
@@ -39,6 +39,11 @@ RUN echo "armhfp" > /etc/yum/vars/basearch && \
 RUN rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
     rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-SIG-AltArch-Arm32
 
+# Centos 7 EOLed on July 1st, 2024. We need to switch to vault.centos.org to get the packages
+RUN sed -i s/mirror.centos.org/vault.centos.org/g /etc/yum.repos.d/*.repo && \
+    sed -i s/^#.*baseurl=http/baseurl=http/g /etc/yum.repos.d/*.repo && \
+    sed -i s/^mirrorlist=http/#mirrorlist=http/g /etc/yum.repos.d/*.repo
```
chouquette, line 45: I'm not entirely sure the shell won't attempt to substitute things here, maybe start with quoting and see if it improve the result before looking for a more complicated cause?

chouquette:
```suggestion
    sed -i 's/^#.*baseurl=http/baseurl=http/g' /etc/yum.repos.d/*.repo && \
    sed -i 's/^mirrorlist=http/#mirrorlist=http/g' /etc/yum.repos.d/*.repo
```

rpm-x64/Dockerfile
```diff
@@ -39,6 +39,11 @@ ENV DD_TARGET_ARCH $DD_TARGET_ARCH
 # persist RHEL major for readable output
 RUN echo $(cat /etc/redhat-release | cut -d'.' -f1 | awk '{print $NF}') > /etc/redhat-release-major
 
+# Centos 7 EOLed on July 1st, 2024. We need to switch to vault.centos.org to get the packages
+RUN sed -i s/mirror.centos.org/vault.centos.org/g /etc/yum.repos.d/*.repo && \
+    sed -i s/^#.*baseurl=http/baseurl=http/g /etc/yum.repos.d/*.repo && \
+    sed -i s/^mirrorlist=http/#mirrorlist=http/g /etc/yum.repos.d/*.repo
```
FlorentClarret, line 45:
```suggestion
    sed -i 's/^#.*baseurl=http/baseurl=http/g' /etc/yum.repos.d/*.repo && \
    sed -i 's/^mirrorlist=http/#mirrorlist=http/g' /etc/yum.repos.d/*.repo
```

chouquette:
> > This looks OK but I was hoping to do the change in a common base image so we don't have to sed in a gazillion Dockerfile ([DataDog/images#5384](https://github.com/DataDog/images/pull/5384))
> 
> Note that if this common base image is in a private registry, then it is not suitable as a base for the images in this public repository, which should be buildable without access to our internal infrastructure.

Ah good point, I was solely focusing on the image being accessible once built.

rpm-armhf/Dockerfile
```diff
@@ -39,6 +39,11 @@ RUN echo "armhfp" > /etc/yum/vars/basearch && \
 RUN rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
     rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-SIG-AltArch-Arm32
 
+# Centos 7 EOLed on July 1st, 2024. We need to switch to vault.centos.org to get the packages
+RUN sed -i s/mirror.centos.org/vault.centos.org/g /etc/yum.repos.d/*.repo && \
```
KevinFairise2, line 43: I think the used mirror is `mirrorlist.centos.org`

chouquette: Indeed, but this URL is only used in the `mirrorlist` section that gets disabled

The other required change (AFAIK) is to replace `baseurl=mirror.centos.org` by `baseurl=vault.centos.org`

KevinFairise2: Ah yes you're right!

KevinFairise2 (approved):

CelianR (approved):

PR #619, opened 2024-07-02
btf-gen: clone the correct submodule version

chouquette:
The current code clones the submodule as defined in main, only to switch to another version afterward.
This makes the bpftools and libbpf code out of sync, which now becomes apparent as libbpf removed an API that is still used in the pinned commit: https://gitlab.ddbuild.io/DataDog/datadog-agent-buildimages/-/jobs/559017010


gjulianm (approved):
Thanks for the fix! I was trying to update to the latest bpftool version but this is far quicker :D

PR #626, opened 2024-07-26
deb-x64: install jq

chouquette:
We want to attempt a fix for the flakyness in OCI packaging jobs that involves URL encoding the CI_JOB_TOKEN we use, and we'd rather use `jq` that doing so in plain bash


FlorentClarret (approved):

FlorentClarret:
Discussed with Hugo, closing for now!

PR #629, opened 2024-07-31
place python before ruby on PATH to deconflict aws cli

clarkb7:
Changes
---
Windows only

Install Python before Ruby so it is first on User PATH

Pass `nomodpath` to Ruby installer to prevent it from putting ruby on the Machine PATH

Motivation
---
Effort to remove hardcoded python install path in [Agent omnibus build](https://github.com/DataDog/datadog-agent/blob/ceaa7b1c03c75d77cac46c01374ad74db8a88738/omnibus/config/software/datadog-agent-integrations-py3.rb#L135-L136)


clarkb7:
closing.

Our build depends on `ridk enable` which puts ruby back at the front of the PATH anyway, so this won't help.

PR #641, opened 2024-08-21
Move from upstream go to microsoft go

Kaderinho:
Install Microsoft Go as well and use GOROOT to select which toolchain to use. The main motivation is to use Microsoft Go to enable FIPS mode in the Agent.

Also, there is a PR to add a [nightly pipeline to create buildimages](https://github.com/DataDog/gitlab-config/pull/165/files) that uses Go Upstream. Note since both toolchain are installed, GOROOT can be changed so we technically can compile using Go upstream or Microsoft Go with a same buildimage.


go.env
pgimalac, line 1: Can you check if the update go invoke task overrides the whole file or if it cleanly updates only the values it needs to ?

Kaderinho: oh actually this reminds me to add msgo update to the invoke task, thanks

pgimalac: Yep that works too !

circleci/Dockerfile
```diff
@@ -50,12 +50,23 @@ RUN set -ex \
     rm -rf /var/lib/apt/lists/*
 
 # Golang
+ARG GO_TOOLCHAIN
+ARG MSGO_VERSION
+ARG MSGO_SHA256_LINUX_AMD64
 ARG GO_VERSION
 ARG GO_SHA256_LINUX_AMD64
+ENV MSGO_VERSION $MSGO_VERSION
```
chouquette, line 58: Is this variable needed in the container? (Same question for other occurrences of the
```
ARG FOO
ENV FOO $FOO
```
pattern later)

circleci/Dockerfile
```diff
@@ -50,12 +50,23 @@ RUN set -ex \
...
-    && echo "$GO_SHA256_LINUX_AMD64 /tmp/golang.tar.gz" | sha256sum --check \
+RUN if [ "$GO_TOOLCHAIN" = "google" ]; then \
+        echo "Installing upstream Go" \
+        && curl -sL -o /tmp/golang.tar.gz https://go.dev/dl/go$GO_VERSION.linux-amd64.tar.gz \
+        && echo "$GO_SHA256_LINUX_AMD64  /tmp/golang.tar.gz" | sha256sum --check; \
+    else \
+        echo "Installing Microsoft Go" \
+        && curl -SL -o /tmp/golang.tar.gz https://aka.ms/golang/release/latest/go${MSGO_VERSION}.linux-amd64.tar.gz \
+        && echo "$MSGO_SHA256_LINUX_AMD64  /tmp/golang.tar.gz" | sha256sum --check; \
+    fi \
```
chouquette, line 69: Kind of a nit, but wouldn't it be more readable to dispatch the appropriate shasum & url based on `GO_TOOLCHAIN` and keep the same script?
Something like
```
GO_URL=go-upstream
GO_CHECKSUM=...
if [ "$GO_TOOLCHAIN" != "google ]; then
    GO_URL=aka.ms/...
    GO_CHECKSUM=...
fi
echo "Installing $GO_TOOLCHAIN"
curl $GO_URL
shasum $GO_CHECKSUM ...
```

PR #642, opened 2024-08-22
Update cmake version

alopezz:
We use CMake mainly to build rtloader. CMake's functionality around finding Python hardcodes the possible Python version values, and the version we are currently using has it hardcoded up to 3.11. That causes our[ initial attempt](https://github.com/DataDog/datadog-agent/pull/28492) at bumping the Agent's embedded Python version to 3.12 to [fail](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/609430786#L2854).

Version 3.30.2 includes support for finding versions up to 3.13: https://gitlab.kitware.com/cmake/cmake/-/blob/v3.30.2/Modules/FindPython/Support.cmake?ref_type=tags#L33.


chouquette (approved):

alopezz:
I've done a test run https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/621749811 that confirms that it finds the correct Python (and we know it's not the system python because it's not flagged by the health check).

PR #644, opened 2024-08-26
Allow publishing using registry.ddbuild.io

KSerrania:
Adds a template job to publish images to registry.ddbuild.io instead of ECR for internal use. This will allow migrating the buildimages from ECR to registry.ddbuild.io.

This uses `--label target=none`, since the images are only used in CI, and do not need to be replicated.


spencergilbert:
> This uses --label target=none, since the images are only used in CI, and do not need to be replicated

Noting this requires the downstream consumers of the image to update the registry from ECR -> dd-registry

chouetz (approved):

PR #656, opened 2024-09-10
[CI] Add the `contents:write` permission to the `go-update.yml` workflow

amenasria:
The `go-update.yml` workflow needs the `contents:write` permission to be able to commit and create a PR on the repository. Else we get this error ([see job](https://github.com/DataDog/datadog-agent-buildimages/actions/runs/10792379054/job/29931896062)):
```
RequestError [HttpError]: GitHub Actions is not permitted to create or approve pull requests.
```
This PR fixes it.

### DIFF
--- a/.github/workflows/go-update.yml
+++ b/.github/workflows/go-update.yml
@@ -23,11 +23,6 @@ on:
         type: string
         default: ''
 
-permissions:
-  # Give the default GITHUB_TOKEN write permission to commit and push the
-  # added or changed files to the repository.
-  pull-requests: write
-
 env:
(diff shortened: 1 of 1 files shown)


.github/workflows/go-update.yml
```diff
@@ -23,7 +23,10 @@ on:
         type: string
         default: ''
 
-permissions: {}
+permissions:
```
chouetz, line 26: If I'm correct that's set just below on the job at lines 37 to 39 for the whole job (contents: write to create branch and pull-requests:write to write into the PR)

chouetz:
I think we can close this PR

PR #664, opened 2024-09-13
Add new linux image with a custom toolchain targeting a specific glibc version

chouquette:
The intent is to stop relying on the oldest OSes we need to support to build the agent,

Instead, we can build a toolchain that will target the appropriate glibc versions (currently, glibc 2.17 for x86_64 and 2.23 for arm64)

This requires some changes in the agent repo, and in both omnibus repos as well.

This currently doesn't include an `armhf` image, which we could use for the IOT agent. We can also not care and stick to whatever we're currently doing, ie keep using a centos 7 or similar image when building for armhf.
We might also be able to include support for armhf into the arm image and cross compile from arm64 to armhf, but this might be much more work for not much gain.

Related to https://github.com/DataDog/datadog-agent/pull/28262


linux-glibc-2.17-x64/Dockerfile
```diff
@@ -0,0 +1,117 @@
...
+
+RUN git config --global user.email "package@datadoghq.com"
+RUN git config --global user.name "Bits"
+
+# Build texinfo to get a bugfix version
+# Otherwise generating glibc doc with a recent version
+# causes a segfault. The fix isn't yet released as a package
+# https://git.savannah.gnu.org/cgit/texinfo.git/commit/?id=60d3edc4b74b4e1e5ef55e53de394d3b65506c47
+RUN cd /build && \
+    git clone --depth=1 https://git.savannah.gnu.org/git/texinfo.git && \
```
f4usto, line 40: **suggestion**:
Can we checkout to a specific revision/tag? Seems a bit risky to rely on the main HEAD

linux-glibc-2.17-x64/Dockerfile
```diff
@@ -0,0 +1,117 @@
...
+
+RUN cd /build && \
+    wget http://crosstool-ng.org/download/crosstool-ng/crosstool-ng-${CTNG_VERSION}.tar.xz && \
+    gpg --keyserver pgp.surfnet.nl --recv-keys 1F30EF2E && \
+    wget http://crosstool-ng.org/download/crosstool-ng/crosstool-ng-${CTNG_VERSION}.tar.xz.sig && \
+    gpg --verify crosstool-ng-${CTNG_VERSION}.tar.xz.sig && \
+    tar xf crosstool-ng-${CTNG_VERSION}.tar.xz && \
+    cd /build/crosstool-ng-${CTNG_VERSION} && \
+    ./configure --enable-local && make -j$(nproc) && \
+    export CT_ALLOW_BUILD_AS_ROOT_SURE=yes && \
```
f4usto, line 54: if we need that set, should we set outside the RUN command or unset later?

f4usto (commented):
Thanks for the effort @chouquette 

Can you provide more details on the reason for maintaining 2.17 and 2.23 instead of unifying them on 2.17. On the original RFC, I don't recall that mentioned.

On the `armhf` topic, we probably understand better the effort not only of adding the x-compile, but the expected maintenance vs the old OS image. IMHO , we can keep it separated fornow

linux-glibc-2.17-x64/config.toml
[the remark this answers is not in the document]
chouquette: It is rust/cargo specific, I renamed the file to `cargo-config.toml`

linux-glibc-2.17-x64/Dockerfile
```diff
@@ -0,0 +1,117 @@
...
+# Remove the early return on non-interactive shells, which makes sourcing the file not activate conda
+RUN grep -v return /root/.bashrc >> /root/newbashrc && cp /root/newbashrc /root/.bashrc
+
+RUN apt update -qy && apt install -y \
+    wget xz-utils gpg build-essential flex texinfo unzip \
+    help2man file gawk libtool-bin bison libncurses-dev \
+    python-is-python3 git cmake curl fakeroot procps bzip2 \
+    pkg-config libssl-dev libcurl4-openssl-dev libexpat-dev libpq-dev libz-dev \
+    rpm tar gettext autopoint autoconf clang libtool-bin \
+    pkg-config flex meson selinux-basics squashfs-tools gpg xz-utils gnupg2 patchelf cpio
```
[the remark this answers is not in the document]
chouquette: That feels like a leftover from testing, I'll remove `jq` as I can't say what it's used for (worst case, the CI will remind me :sweat_smile: )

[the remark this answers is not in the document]
chouquette: Oh no wait, `jq` isn't there. Where you somehow referring to https://github.com/DataDog/datadog-agent-buildimages/pull/626?

linux-glibc-2.17-x64/Dockerfile
chouquette: Fair enough, I pinned the commit to the one that fixes our issue

linux-glibc-2.17-x64/Dockerfile
chouquette: It only matters while building the toolchain, not when using it, so it doesn't need to be set outside of this specific command

.gitlab-ci.yml
```diff
@@ -231,6 +231,18 @@ build_dd_agent_testing:
     DOCKERFILE: dd-agent-testing/Dockerfile
     IMAGE: dd-agent-testing
 
+build_linux_glibc_2_17_x64:
+  extends: [.build, .x64]
+  variables:
+    DOCKERFILE: linux-glibc-2.17-x64/Dockerfile
```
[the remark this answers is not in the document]
chouquette: We could definitely talk about grouping those in a `linux` folder and have the permutations split by folders

We don't need to have images for deb/rpm anymore, anything can be built on one or the other, glibc requirement aside, but with these images this requirement goes away.

The name isn't set in stone and I'll happily take suggestions!

linux-glibc-2.17-x64/Dockerfile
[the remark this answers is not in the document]
spencergilbert: Probably referring to the change in [deb-x64/Dockerfile](https://github.com/DataDog/datadog-agent-buildimages/pull/664/files?diff=unified&w=0#diff-a2b47c4829b01f98d0ce2dcb11c6e5398adef83123aa1b1d5c2c9d25d3bb8292R62)?

linux-glibc-2.17-x64/Dockerfile
```diff
@@ -0,0 +1,118 @@
...
+# https://git.savannah.gnu.org/cgit/texinfo.git/commit/?id=60d3edc4b74b4e1e5ef55e53de394d3b65506c47
+RUN cd /build && \
+    git clone https://git.savannah.gnu.org/git/texinfo.git && \
+    cd texinfo && \
+    git reset --hard 60d3edc4b74b4e1e5ef55e53de394d3b65506c47 && \
+    ./autogen.sh && \
+    ./configure && make -j$(ncproc) && make install && \
+    rm -rf /build/texinfo
+
+RUN cd /build && \
```
spencergilbert, line 47: **❓ question:** ‏ are these not available at `https` addresses?

linux-glibc-2.17-x64/Dockerfile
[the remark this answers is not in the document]
chouquette: Oh I completely missed this change, that's definitely not intended and will be removed

linux-glibc-2.17-x64/Dockerfile
chouquette: They are not, or not in a reliable fashion. Currently it times out when I thy to access that URL through HTTPS.

You might want to ask what's the point of checking a GPG signature that's downloaded over HTTP, and you'd be right to, however I have no answer :(

spencergilbert: Weird - but fine

chouquette: Heh I guess that it's not possible to forge a signature file that would still match the private key, and the key itself is delivered over a secured channel

KSerrania: For the sake of not having to use HTTP, could we use the artifacts from GitHub releases instead?

They provide the `.tar.xz` and the signature file there as well: https://github.com/crosstool-ng/crosstool-ng/releases/tag/crosstool-ng-1.26.0

linux-glibc-2.17-x64/Dockerfile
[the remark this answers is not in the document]
chouquette: push forced into oblivion

linux-glibc-2.17-x64/Dockerfile
spencergilbert: 🤔 would an `export`ed variable get included in the final image? I'm not sure off the top of my head.

chouquette: No, this variable will stop existing outside of this specific layer

linux-glibc-2.17-x64/Dockerfile
chouquette: Very good point, done!

chouetz (approved):
LGTM

f4usto (approved):
lgtm, thanks for the diligence on addressing the comments

PR #665, opened 2024-09-17
Chouquette/restore working images

chouquette:
This PR reverts our build images to a working state by reverting all cmake & python changes, and a windows related change which appears to break the windows build consistently.

cc @Kyle-Neale @alopezz @julien-lebot as your PR are part of the reverted ones.

The main motivation here is that we fairly urgently need to be able to bump golang's version and we currently can't as the top of main doesn't provide us with working images


alopezz (approved):
👍 As discussed I hope we can recover and test the CMake update as soon as we can

PR #677, opened 2024-09-24
upgrade perl's List::Util module

chouquette:
OpenSSL's build system requires 1.29+ and we currently ship 1.27

### DIFF
--- a/deb-x64/Dockerfile
+++ b/deb-x64/Dockerfile
@@ -68,6 +68,10 @@ RUN apt-get update \
     && apt-get install -y gcc-9 g++-9 \
     && update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-9 0
 
+# Upgrade some perl modules since some of our dependencies require some "new" features
+# that were not available in centos 7
+RUN cpanp -i List::Util
+
 # Ensure we're using the newest binutils
 ENV PATH "/usr/lib/binutils-2.26/bin:$PATH"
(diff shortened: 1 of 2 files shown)


chouquette:
Tested with the OpenSSL update together in this pipeline: https://gitlab.ddbuild.io/DataDog/datadog-agent/-/pipelines/45004004

It seems we'll also need to update the module in the rpm arm64 image

deb-x64/Dockerfile
```diff
@@ -68,6 +68,10 @@ RUN apt-get update \
     && apt-get install -y gcc-9 g++-9 \
     && update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-9 0
 
+# Upgrade some perl modules since some of our dependencies require some "new" features
+# that were not available in centos 7
+RUN cpanp -i List::Util
```
alopezz, line 73: Question: would it be possible to pin to a version?

chouquette: Good point. done!

alopezz (approved):

f4usto (approved):

chouquette:
new attempt: https://gitlab.ddbuild.io/DataDog/datadog-agent/-/pipelines/45022171

chouquette:
Build seems to be ok with this change and the openSSL bump, merging the build image and I'll update the PR tomorrow

PR #678, opened 2024-09-24
Use same GO version in agent-deploy as in other images

KevinFairise2:
Use the same GO version for `agent-deploy` image as in the other buildimages.

Want to update that because otherwise I am not able to efficiently use the cache on datadog-agent. Cache does not work with different version of Go

### DIFF
--- a/agent-deploy/Dockerfile
+++ b/agent-deploy/Dockerfile
@@ -13,8 +13,8 @@ ARG RUBY_VERSION_ARG=2.4.10
 ARG BUNDLER_VERSION_ARG=1.17.3
 ARG AWSCLI_VERSION_ARG=1.11.18
 ARG RPM_S3_VERSION_ARG=0.8.0
-ARG GOLANG_VERSION=1.21.7
-ARG GOLANG_SHASUM=13b76a9b2a26823e53062fa841b07087d48ae2ef2936445dc34c4ae03293702c
+ARG GO_VERSION
+ARG GO_SHA256_LINUX_AMD64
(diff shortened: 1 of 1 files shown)


chouquette (approved):

agent-deploy/Dockerfile
```diff
@@ -13,8 +13,8 @@ ARG RUBY_VERSION_ARG=2.4.10
 ARG BUNDLER_VERSION_ARG=1.17.3
 ARG AWSCLI_VERSION_ARG=1.11.18
 ARG RPM_S3_VERSION_ARG=0.8.0
-ARG GOLANG_VERSION=1.21.7
-ARG GOLANG_SHASUM=13b76a9b2a26823e53062fa841b07087d48ae2ef2936445dc34c4ae03293702c
+ARG GO_VERSION
+ARG GO_SHA256_LINUX_AMD64
```
chouetz, line 17: Seems like [most of](https://github.com/search?q=repo%3ADataDog%2Fdatadog-agent-buildimages%20GO_VERSION&type=code) the other images have [this](https://github.com/DataDog/datadog-agent-buildimages/blob/ff4a2bf38f471c23d3c1c5954a25aa9e101cdb0e/rpm-arm64/Dockerfile#L35)
```
ENV GO_VERSION $GO_VERSION
```
I'm not sure if this is required, for the sake of uniformity it could be nice to have it, wdyt?

chouetz (approved):

agent-deploy/Dockerfile
KevinFairise2: Why not, we currently do not use it. But we can add it

PR #682, opened 2024-09-25
feat(readme): Add a badge with the latest image version published from `main`

chouetz:
Check the result [here](https://github.com/DataDog/datadog-agent-buildimages/blob/nschweitzer/badge/README.md)
Add the latest version published easy to see.
Copy of it seems not trivial (you have to click on the image then select)


amenasria (approved):
Really nice !!

PR #683, opened 2024-09-25
Fix image releases

chouquette:
A recent refactoring combined with a counter intuitive gitlab behavior was causing our image release jobs to systematically fail.
This PR fixes the variables we provide to `public-images` jobs, which should in turn allow us to publish our circle CI images

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -527,7 +527,7 @@ release_circleci_runner_gcr:
     - !reference [.on_default_branch_push]
   variables:
     IMG_SOURCES: datadog/agent-buildimages-circleci-runner${ECR_TEST_ONLY}:$IMAGE_VERSION
-    IMG_DESTINATIONS: agent-circleci-runner:$IMAGE_VERSION
+    IMG_DESTINATIONS: agent-circleci-runner${ECR_TEST_ONLY}:$IMAGE_VERSION
     IMG_REGISTRIES: gcr-datadoghq
   trigger:
     project: DataDog/public-images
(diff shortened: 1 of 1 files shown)


chouetz (approved):

.gitlab-ci.yml
```diff
@@ -526,8 +531,13 @@ release_circleci_runner_gcr:
   rules:
     - !reference [.on_default_branch_push]
   variables:
+    # Don't rely on ${IMAGE_VERSION} or any variables that is referencing
+    # other variables here.
+    # They are not expended recursively and end up provided verbatim to the
+    # downstream pipeline which then extends the variable with its own CI
+    # variables, which causes invalid image tags to be used
     IMG_SOURCES: datadog/agent-buildimages-circleci-runner${ECR_TEST_ONLY}:$IMAGE_VERSION
```
KSerrania, line 539: I think this line also needs an update?

.gitlab-ci.yml
```diff
@@ -526,8 +531,13 @@ release_circleci_runner_gcr:
   rules:
```
KSerrania, line 531: Note: with this change, `dev_release_circleci_runner_gcr` and `release_circleci_runner_gcr` are identical. That's strange, we should check why that is the case.

chouquette: It *seems* that the intent was to somehow replicate the '${TEST_ONLY}` behavior, by having one job for the default branch and one job for other pushes (also including `main` FWICS)
This can probably be merged in a single job, but I'd prefer to do so once we're back to a working state (and quite possibly with more thorough verifications)

.gitlab-ci.yml
chouquette: At least I added the comment to make it clear that I forgot to update :sob:

PR #693, opened 2024-10-02
use build_ddregistry base job for new glibc toolchains

chouquette:
The non test_only ECR was cleaned up due to not being used in a while, which causes failures on main
This feels like a good opportunity to start deploying through registry.ddbuild.io

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -252,7 +252,7 @@ build_linux_glibc_2_17_x64:
     IMAGE: linux-glibc-2-17-x64
 
 build_linux_glibc_2_23_arm64:
-  extends: build_ddregistry .arm]
+  extends: [.build_ddregistry, .arm]
   variables:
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -246,13 +246,13 @@ build_dd_agent_testing:
...
 build_linux_glibc_2_17_x64:
-  extends: [.build, .x64]
+  extends: [.build_ddregistry, .x64]
   variables:
     DOCKERFILE: linux-glibc-2.17-x64/Dockerfile
     IMAGE: linux-glibc-2-17-x64
 
 build_linux_glibc_2_23_arm64:
-  extends: [.build, .arm]
+  extends: build_ddregistry .arm]
```
KSerrania, line 255:
```suggestion
  extends: [.build_ddregistry, .arm]
```

chouquette: Indeed :facepalm: Thanks!

PR #695, opened 2024-10-03
feat(rpm): Add vault in other images

chouetz:
I forgot some images leading to [failed pipeline](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/659673375)


CelianR (approved):

spencergilbert (approved):
🤔 I think we could reduce the duplication across dockerfiles and use `COPY --from` to include the `vault` binary from a shared image definition.

chouetz:
> 🤔 I think we could reduce the duplication across dockerfiles and use `COPY --from` to include the `vault` binary from a shared image definition.

I fully agree that there are several tools that can be shared (I tried to install only on the required images, but missed that one.
I'll prepare some thoughts about improvements in buildimages for early '25

PR #702, opened 2024-10-15
Add developer environment for Linux

ofek:
This adds our new standard developer environment for Linux which will be used by our [tooling](https://github.com/DataDog/datadog-agent-dev).

Notes:

- There will be a single image for Linux rather than per architecture
- Publishing via CI is not yet set up
- This Visual Studio Code setup will "supposedly" stop working in February as mentioned in `dev-envs/linux/vscode.sh` because they will drop support for connecting via SSH to older machines. Therefore we must build from source. It requires a modern version of NodeJS which also is unsupported on our very old build image so we must actually build that from source first in order to build Visual Studio Code


chouquette:
> they will drop support for connecting via SSH to older machines. Therefore we must build from source. It requires a modern version of NodeJS which also is unsupported on our very old build image so we must actually build that from source first in order to build Visual Studio Code

This sounds like a good use case to start using our recently introduced build images based on ubuntu 24.04 which still support building targeting glibc 2.17 or 2.23 depending on the architecture.

A minor issue is that their name depends on the glibc version, not only the arch, so it will require a bit more plumbing.
They are named `linux-glibc-2.17-x64` and `linux-glibc-2.23-arm64` if you want to give it a try

ofek:
Unfortunately the minimum requirement is 2.28 https://code.visualstudio.com/docs/supporting/FAQ#_can-i-run-vs-code-on-older-linux-distributions

chouquette:
> Unfortunately the minimum requirement is 2.28 https://code.visualstudio.com/docs/supporting/FAQ#_can-i-run-vs-code-on-older-linux-distributions

These images are running on glibc 2.39 (but provide a toolchain that builds for 2.17 or 2.23), so I think that should be OK (unless I misunderstood what this does)

vickenty:
Would it make sense to place this under the existing `devenv` directory, rather than introducing another top-level one with just a slightly different name?

ofek:
@vickenty Do you mean `devcontainer`?

image

vickenty:
@ofek No, I meant [this](https://github.com/DataDog/datadog-agent/tree/main/devenv), but now I realized it's a different repository! Please ignore me.

dev-envs/linux/ssh.sh
```diff
@@ -0,0 +1,40 @@
...
+make install
+popd
+rm -rf "${workdir}"
+
+# Ameliorate transient network issues for things that can
+# retry like Visual Studio Code's Remote - SSH extension
+passwd -d root
+cat <<'EOF' >> /usr/local/etc/sshd_config
+PermitRootLogin yes
+PermitEmptyPasswords yes
```
pducolin, line 30: **💬 suggestion**
An alternative to this is setting up authenticated_keys passing a public key at devenv creation. This way when you first create an instance of the container, it is setup to accept your ssh key connection

dev-envs/linux/ssh.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash -l
+IFS=$'\n\t'
+set -euxo pipefail
+
+# New versions require OpenSSL >= 1.1.1
+VERSION="V_9_3_P2"
+url="https://github.com/openssh/openssh-portable/archive/refs/tags/${VERSION}.tar.gz"
```
pducolin, line 7: **❓ question**
Why not installing it from apt?

dev-envs/linux/ssh.sh
ofek: What is the benefit if it's on your local machine or in the cloud behind AppGate?

.gitlab/dev_envs.yml
```diff
@@ -0,0 +1,91 @@
...
+  image: "$BUILDENV_REGISTRY/images/docker:20.10-py3"
+  variables:
+    BASE_IMAGE_TAG: v47431157-7a59834c
+    BASE_IMAGE_ARCH: arm64
+    TARGET_IMAGE_ARCH: arm64
+
+.build_dev_env:
+  script:
+    # Docker Hub login
+    - DOCKER_REGISTRY_LOGIN=$(aws ssm get-parameter --region us-east-1 --name ci.datadog-agent.docker_hub_login --with-decryption --query "Parameter.Value" --out text)
```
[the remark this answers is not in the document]
ofek: Sure, I'll look for an example!

.gitlab/dev_envs.yml
```diff
@@ -0,0 +1,91 @@
...
+  needs: [build_dev_env_linux_amd64, build_dev_env_linux_arm64]
+  rules:
+    - if: '$CI_COMMIT_BRANCH == "main"'
+  trigger:
+    project: DataDog/public-images
+    branch: main
+    strategy: depend
+  variables:
+    IMG_SOURCES: registry.ddbuild.io/ci/datadog-agent-buildimages/dev-env-linux:amd64,registry.ddbuild.io/ci/datadog-agent-buildimages/dev-env-linux:arm64
+    IMG_DESTINATIONS: agent-dev-env-linux:latest
```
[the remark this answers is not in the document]
ofek: Done, thanks!

dev-envs/linux/git.sh
```diff
@@ -0,0 +1,44 @@
...
+IFS=$'\n\t'
+set -euxo pipefail
+
+VERSION="2.37.3"
+url="https://github.com/git/git/archive/refs/tags/v${VERSION}.tar.gz"
+
+archive_name=$(basename "${url}")
+workdir="/tmp/setup-${archive_name}"
+mkdir -p "${workdir}"
+curl "${url}" -Lo "${workdir}/${archive_name}"
```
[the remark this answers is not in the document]
ofek: Done! I didn't check the hashes of the fonts because that didn't seem too useful.

dev-envs/linux/scripts/git-dd-clone.sh
```diff
@@ -0,0 +1,31 @@
...
+        set +x
+        cd "${clone_dir}"
+        set -x
+
+        git switch -c "${branch}"
+    fi
+fi
+
+# The first time is a bit slow so do that now for a better experience
+gfold "${clone_dir}" > /dev/null 2>&1
```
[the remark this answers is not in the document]
ofek: The first time before caching can be a bit slow so we do it here so users who wish to work with multiple repositories using that command have a nice, responsive experience. If you want I can remove it, let me know.

.gitlab/dev_envs.yml
```diff
@@ -0,0 +1,91 @@
...
+    # Build
+    - docker build dev-envs/$PLATFORM --build-arg ARCH=$BASE_IMAGE_ARCH --build-arg TAG=$BASE_IMAGE_TAG
+
+.publish_dev_env:
+  script:
+    # Docker Hub login
+    - DOCKER_REGISTRY_LOGIN=$(aws ssm get-parameter --region us-east-1 --name ci.datadog-agent.docker_hub_login --with-decryption --query "Parameter.Value" --out text)
+    - aws ssm get-parameter --region us-east-1 --name ci.datadog-agent.docker_hub_pwd --with-decryption --query "Parameter.Value" --out text | docker login --username "$DOCKER_REGISTRY_LOGIN" --password-stdin docker.io
+    # Build and push
+    - docker build dev-envs/$PLATFORM --build-arg ARCH=$BASE_IMAGE_ARCH --build-arg TAG=$BASE_IMAGE_TAG --tag registry.ddbuild.io/ci/datadog-agent-buildimages/dev-env-$PLATFORM:$TARGET_IMAGE_ARCH
```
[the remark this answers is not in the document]
ofek: We can't use crane in this area because we have to push a multi-architecture image in the end. To your point however I was erroneously extending a block that was unnecessary so I removed that!

dev-envs/linux/ssh.sh
ofek: The base image is very old so it's easier to get the source directly.

.gitlab/dev_envs.yml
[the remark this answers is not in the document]
ofek: Based on offline discussions, this will wait until a separate PR that migrates the entire repo since Vault is not yet used here at all.

KevinFairise2 (approved):

.gitlab/dev_envs.yml
```diff
@@ -0,0 +1,91 @@
...
+    - docker push registry.ddbuild.io/ci/datadog-agent-buildimages/dev-env-$PLATFORM:$TARGET_IMAGE_ARCH
+
+lint_dev_envs:
+  tags: ["arch:amd64"]
+  stage: dev_envs
+  needs: []
+  image: 486234852809.dkr.ecr.us-east-1.amazonaws.com/ci/images:v26997867-ebc48a76
+  script:
+    - hadolint dev-envs/linux/Dockerfile
```
chouetz, line 40: Maybe for a later PR, I assume we could lint all `Dockerfiles` like this, correct?

dev-envs/linux/Dockerfile
```diff
@@ -0,0 +1,76 @@
...
+COPY env.sh /setup/env.sh
+COPY <<-"EOF" /setup/shellrc.sh
+set -euxo pipefail
+
+export PATH="${HOME}/.scripts:${PATH}"
+EOF
+
+# Set up Zsh
+COPY zsh.sh /setup/zsh.sh
+RUN /setup/zsh.sh
```
chouetz, line 24: Even if it's much more readable/"debuggable" like that, should we have grouped `COPY` or `RUN` commands to have a lighter image?

chouetz (approved):

.gitlab/dev_envs.yml
ofek: Definitely that's possible! Do we hint `Dockerfile`s elsewhere?

ofek:
I'm going to merge this as the error is unrelated.

PR #713, opened 2024-11-07
[Docker-x64] Install Go

amenasria:
This PR installs Go on the docker-x64 build image. 

It is needed for the integration tests to run fine (see https://github.com/DataDog/datadog-agent/pull/30856)

### DIFF
--- a/docker-x64/Dockerfile
+++ b/docker-x64/Dockerfile
@@ -9,7 +9,6 @@ ARG VAULT_VERSION=1.17.2
 ARG VAULT_CHECKSUM=a0c0449e640c8be5dcf7b7b093d5884f6a85406dbb86bbad0ea06becad5aaab8
 ARG VAULT_FILENAME="vault_${VAULT_VERSION}_linux_amd64.zip"
 
-
 RUN apt-get update &&  \
     DEBIAN_FRONTEND=noninteractive apt-get install -y make  \
       build-essential  \
--- a/docker-x64/Dockerfile
+++ b/docker-x64/Dockerfile
@@ -1,11 +1,17 @@
 FROM 486234852809.dkr.ecr.us-east-1.amazonaws.com/docker:20.10
 
 ARG PYTHON_VERSION=3.12.6
+ARG GO_VERSION
+ARG GO_SHA256_LINUX_AMD64
 ARG CI_UPLOADER_VERSION=2.38.1
 ARG CI_UPLOADER_SHA=4e56d449e6396ae4c7356f07fc5372a28999aacb012d4343a3b8a9389123aa38
 ARG VAULT_VERSION=1.17.2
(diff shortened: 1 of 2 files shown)


docker-x64/Dockerfile
```diff
@@ -1,12 +1,18 @@
...
+ARG GO_SHA256_LINUX_AMD64
 ARG CI_UPLOADER_VERSION=2.38.1
 ARG CI_UPLOADER_SHA=4e56d449e6396ae4c7356f07fc5372a28999aacb012d4343a3b8a9389123aa38
 ARG VAULT_VERSION=1.17.2
 ARG VAULT_CHECKSUM=a0c0449e640c8be5dcf7b7b093d5884f6a85406dbb86bbad0ea06becad5aaab8
 ARG VAULT_FILENAME="vault_${VAULT_VERSION}_linux_amd64.zip"
 
+# Environment
+ENV GOPATH /go
+ENV GO_VERSION $GO_VERSION
```
L3n41c, line 14: This is already done further in the same `Dockerfile` line 55 and 56: https://github.com/DataDog/datadog-agent-buildimages/pull/713/files#diff-5dab46fa1b83d99a3786c557458cf2336c672cbb10fbe95dbce71879d05ae715R55

We don’t need to set the same environment variables twice.

```suggestion
```

docker-x64/Dockerfile
```diff
@@ -41,6 +47,16 @@ COPY requirements /requirements
...
 
+# Install Go
+RUN curl -sL -o /tmp/golang.tar.gz https://go.dev/dl/go$GO_VERSION.linux-amd64.tar.gz \
+  && echo "$GO_SHA256_LINUX_AMD64  /tmp/golang.tar.gz" | sha256sum --check \
+  && tar -C /usr/local -xzf /tmp/golang.tar.gz \
+  && rm -f /tmp/golang.tar.gz
+ENV GOPATH /go
+ENV GO_VERSION $GO_VERSION
+ENV PATH="/usr/local/go/bin:${PATH}"
+ENV PATH="${GOPATH}/bin:${PATH}"
```
L3n41c, line 58: Can be done in one statement to save one layers.

```suggestion
ENV PATH="/go/bin:/usr/local/go/bin:${PATH}"
```

L3n41c (commented):
Is it expected that we touch only `docker-x64/Dockerfile` and that we don’t keep `docker-arm64/Dockerfile` in sync ?

docker-x64/Dockerfile
amenasria: Nice catch, thanks !

docker-x64/Dockerfile
amenasria: Good point, thanks !

L3n41c (approved):

CelianR (approved):

PR #715, opened 2024-11-08
Add msgo to buildimages

Kaderinho:
This PR adds [Microsoft go toolchain](https://github.com/microsoft/go) into the `glibc-linux` images so that we can build the Agent with FIPS support.

We're installing msgo along side with upstream go and the selection is done through the `.bashrc` (default: upstream go).
Basically we'll define `DD_GO_TOOLCHAIN="msgo"` to use microsoft GO to build FIPS packages


jeremy-hanna (approved):
datadog-agent ci run with these images: https://gitlab.ddbuild.io/DataDog/datadog-agent/-/pipelines/48606072

(the multi-project ci test run was failing due to [.python-version issues](https://github.com/DataDog/datadog-agent/pull/30332))

setup_go.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+
+set -exo pipefail
```
CelianR, line 3: Is `-x` required? Usually we don't use it to avoid secrets leaks

setup_go.sh
```diff
@@ -0,0 +1,44 @@
...
+
+cat << EOF >> /root/.bashrc
+if [ "\$DD_GO_TOOLCHAIN" = "msgo" ]; then
+    export PATH="/usr/local/msgo/bin:\$PATH"
+    export GOROOT="/usr/local/msgo"
+else
+   export PATH="/usr/local/go/bin:\$PATH"
+   export GOROOT="/usr/local/go"
+fi
+EOF
```
CelianR, line 44: **🥜 nitpick:** ‏Can you add a line feed?

tasks/update_go.py
```diff
@@ -29,30 +29,42 @@ def _get_archive_extension(os: str) -> str:
...
     for os, arch in PLATFORMS:
         ext = _get_archive_extension(os)
-        url = f"https://storage.googleapis.com/golang/go{version}.{os}-{arch}.{ext}.sha256"
+        url = f"{base_url}/go{version}.{os}-{arch}.{ext}.sha256"
         res = requests.get(url)
         res.raise_for_status()
 
-        sha = res.text.strip()
+        # Handle both format "<sha256>" and "<sha256> <filename>"
+        sha_elts = res.text.strip().split("  ")
```
CelianR, line 43: ❓ **question**: Should this be:
```py
        sha_elts = res.text.strip().split(" ")
```
(only one space character within the split)

CelianR: (note that `split()` removes spaces by default)

CelianR (approved):

setup_go.sh
Kaderinho: There should be no secret in this script and this might allow us to better debug go setup. If you still want me do remove it for consistency I can do it 👍

PR #723, opened 2024-12-03
Unzip vault in a dedicated folder

FlorentClarret:
The build jobs for these images fail because we have a LICENSE.txt in `/`, which makes the `unzip` command fail. It seems like we now have vault 1.18.2 in the base image, but I prefer keeping the same version across all our build images. 

https://gitlab.ddbuild.io/DataDog/datadog-agent-buildimages/-/jobs/725118798

Unziping vault in a dedicated folder is a better approach than doing it in `/`

### DIFF
--- a/docker-arm64/Dockerfile
+++ b/docker-arm64/Dockerfile
@@ -54,10 +54,10 @@ ENV PATH="/go/bin:/usr/local/go/bin:${PATH}"
 
 # Install vault: https://github.com/hashicorp/vault/blob/main/CHANGELOG.md https://releases.hashicorp.com/vault
 RUN curl -LO https://releases.hashicorp.com/vault/$VAULT_VERSION/$VAULT_FILENAME \
-    && echo "${VAULT_CHECKSUM}  ${VAULT_FILENAME}" | sha256sum --check \
-    && unzip $VAULT_FILENAME \
+    && echo "${VAULT_CHECKSUM} ${VAULT_FILENAME}" | sha256sum --check \
+    && unzip $VAULT_FILENAME -d /tmp/vault \
     && rm $VAULT_FILENAME \
-    && mv vault /usr/bin/vault \
+    && mv /tmp/vault/vault /usr/bin/vault \
     && chmod +x /usr/bin/vault
 
 # External calls configuration
--- a/docker-x64/Dockerfile
+++ b/docker-x64/Dockerfile
@@ -54,10 +54,10 @@ ENV PATH="/go/bin:/usr/local/go/bin:${PATH}"
 
 # Install vault: https://github.com/hashicorp/vault/blob/main/CHANGELOG.md https://releases.hashicorp.com/vault
 RUN curl -LO https://releases.hashicorp.com/vault/$VAULT_VERSION/$VAULT_FILENAME \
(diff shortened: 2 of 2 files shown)


chouetz (approved):

docker-arm64/Dockerfile
```diff
@@ -54,10 +54,10 @@ ENV PATH="/go/bin:/usr/local/go/bin:${PATH}"
...
 # Install vault: https://github.com/hashicorp/vault/blob/main/CHANGELOG.md https://releases.hashicorp.com/vault
 RUN curl -LO https://releases.hashicorp.com/vault/$VAULT_VERSION/$VAULT_FILENAME \
-    && echo "${VAULT_CHECKSUM}  ${VAULT_FILENAME}" | sha256sum --check \
-    && unzip $VAULT_FILENAME \
+    && echo "${VAULT_CHECKSUM} ${VAULT_FILENAME}" | sha256sum --check \
+    && unzip $VAULT_FILENAME -d /tmp/vault \
     && rm $VAULT_FILENAME \
-    && mv vault /usr/bin/vault \
+    && mv /tmp/vault/vault /usr/bin/vault \
     && chmod +x /usr/bin/vault
```
L3n41c, line 61: Rather than unzipping the whole archive (with the `LICENSE` file we’re not interested in) in a temporary directory and move the file we’re interested in (`vault`) in its final destination in a second step, it’s possible to tell `unzip` to inflate only the file we’re interested in (`vault`) directly in the correct final directory in a single line.

```suggestion
    && unzip -o $VAULT_FILENAME -d /usr/bin -o vault \
    && rm $VAULT_FILENAME
```

docker-x64/Dockerfile
```diff
@@ -54,10 +54,10 @@ ENV PATH="/go/bin:/usr/local/go/bin:${PATH}"
...
 # Install vault: https://github.com/hashicorp/vault/blob/main/CHANGELOG.md https://releases.hashicorp.com/vault
 RUN curl -LO https://releases.hashicorp.com/vault/$VAULT_VERSION/$VAULT_FILENAME \
-    && echo "${VAULT_CHECKSUM}  ${VAULT_FILENAME}" | sha256sum --check \
-    && unzip $VAULT_FILENAME \
+    && echo "${VAULT_CHECKSUM} ${VAULT_FILENAME}" | sha256sum --check \
+    && unzip $VAULT_FILENAME -d /tmp/vault \
     && rm $VAULT_FILENAME \
-    && mv vault /usr/bin/vault \
+    && mv /tmp/vault/vault /usr/bin/vault \
     && chmod +x /usr/bin/vault
```
L3n41c, line 61: As above, we can tell `unzip` to put directly the `vault` file at the right location without inflating `LICENSE` in a single line:

```suggestion
    && unzip -o $VAULT_FILENAME -d /usr/bin -o vault \
    && rm $VAULT_FILENAME
```

docker-arm64/Dockerfile
FlorentClarret: Thanks! Will do this for all the docker files in a follow-up PR 🙇

L3n41c (approved):

PR #736, opened 2025-01-10
Use deva to install dependencies

ofek:
### What does this PR do?

This utilizes deva's lock file to install dependencies rather than using requirement files.

### Motivation

Avoid dependency hell and ensure reproducibility using a single source of truth.

### Additional Notes

* the version of deva is stored in `deva.env`
* in a future PR we will remove Python 2 logic, I didn't touch anything related to that, at which point we can remove the requirement files


setup_python.sh
```diff
@@ -66,7 +66,6 @@ case $DD_TARGET_ARCH in
 
     python3 -m pip install distro==1.4.0 wheel==0.40.0
     python3 -m pip install --no-build-isolation "cython<3.0.0" PyYAML==5.4.1
-    python3 -m pip install -r requirements.txt
```
ofek, line 69: Oversight, should have been part of this https://github.com/DataDog/datadog-agent-buildimages/pull/735

.gitlab-ci.yml
spencergilbert, line 1: **❓ question:** ‏ doublechecking because the diff is miserable, the build command has just been split across lines and the "change" is getting the build args from the deva file?

deva.env
```diff
@@ -0,0 +1,2 @@
+DEVA_VERSION=v0.4.0
+DEVA_NO_DYNAMIC_DEPS=1
```
spencergilbert, line 2: **❓ question:** ‏ is this used elsewhere, since we seem to be setting it "by hand" in each of the dockerfiles?

spencergilbert (approved):

deva.env
ofek: This is used by `deva` itself to indicate required optional dependencies are already available rather than dynamically installing at runtime.

chouetz (approved):

gjulianm (approved):

KevinFairise2 (approved):

PR #738, opened 2025-01-14
Use deva to install dependencies on Windows

ofek:
### What does this PR do?

Final half of https://github.com/DataDog/datadog-agent-buildimages/pull/736

### DIFF
--- a/windows/install-all.ps1
+++ b/windows/install-all.ps1
@@ -27,6 +27,13 @@ foreach ($line in $lines) {
     [Environment]::SetEnvironmentVariable($key, $val, [System.EnvironmentVariableTarget]::Process)
 }
 
+# Read deva variables from deva.env file
+$lines = Get-Content -Path '..\deva.env'
+foreach ($line in $lines) {
+    $key, $val = $line.split('=')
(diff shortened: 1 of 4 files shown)


windows/install-all.ps1
```diff
@@ -95,7 +102,7 @@ try {
     }
 
     if ($Phase -eq 0 -or $Phase -eq 4) {
-        
+
     }
```
CelianR, line 106: **❓ question:** ‏Should this be removed?

CelianR (approved):

windows/install-all.ps1
ofek: Should what be removed?

ofek: Oh I see what you mean, whitespace is insignificant here and this change is fine AFAIK.

clarkb7 (approved):
I'm not sure we should be installing deps into the embedded python but that's what it's doing now so lets leave it.

I think you should also use deva to install deps on the system python. See [install_python.ps1](https://github.com/DataDog/datadog-agent-buildimages/blob/main/windows/helpers/phase2/install_python.ps1)

the embedded python is just used as the python env for unit tests. the system python is used for invoke/etc.

ofek:
Nice catch, thanks!

PR #739, opened 2025-01-16
Remove Python 2

ofek:
### Motivation

We no longer require Python 2

### Additional Notes

A subsequent PR will remove `requirements-py2.txt` as other repositories might reference that still


windows/helpers/phase1/install_vcpython.ps1
```diff
@@ -1,21 +0,0 @@
-
-
-$vcpmsi = 'https://s3.amazonaws.com/dd-agent-omnibus/VCForPython27.msi'
```
julien-lebot, line 3: **❓ question:** ‏ Should we delete that file in `dd-agent-omnibus` as well?

julien-lebot (approved):

windows/helpers/phase1/install_vcpython.ps1
ofek: I don't think we would ever be able to because we would need the ability to build older versions of the agent.

julien-lebot: Makes sense 👍

KevinFairise2 (approved):

chouetz (approved):

PR #744, opened 2025-01-17
Revert "Remove Python 2 (#739)"

chouetz:
### What does this PR do?
This reverts commit c6b0e647fada229d21d642958c51b7a966437c90.

### Motivation
We need python2 for the `rpm-s3` on the `agent-deploy` images


julien-lebot (approved):

ofek (changes requested):
Before reverting can you please link what is required? I would much prefer to upgrade to a newer Python rather than keeping this.

ofek (approved):
Thanks!

usamasaqib (approved):

PR #764, opened 2025-02-18
[ACIX-487] Add linter tooling (shellcheck)

CelianR:
### What does this PR do?

Adds `shellcheck` to images to be used in https://github.com/DataDog/datadog-agent/pull/34099.


chouetz (approved):
Are these the only places where you will need this?
Otherwise LGTM, thanks!

PR #767, opened 2025-02-25
Upgrade developer tooling

ofek:
### Motivation

https://github.com/DataDog/datadog-agent-dev/blob/main/CHANGELOG.md#050---2025-02-25

### Additional Notes

The `deva.env` will be removed in a subsequent PR when other repositories no longer use it


.github/workflows/go-update.yml
pgimalac, line 1: I tried running the workflow and it failed with an error mentioning `dda`
https://github.com/DataDog/datadog-agent-buildimages/actions/runs/13529277004/job/37807333806

pgimalac (changes requested):

.github/workflows/go-update.yml
```diff
@@ -44,13 +44,13 @@ jobs:
       - name: Setup Python and pip
         uses: actions/setup-python@42375524e23c412d93fb67b49958b491fce71c38 # v5.4.0
         with:
-          python-version: 3.x
+          python-version: "3.12"
```
ofek, line 47: Required because the latest major version of Python was bumped to 3.13 and the developer tooling, currently, requires 3.12.

.github/workflows/go-update.yml
ofek: https://github.com/DataDog/datadog-agent-buildimages/pull/767#discussion_r1971577177

.github/workflows/go-update.yml
pgimalac: Thanks !

pgimalac (approved):
I ran the workflow and it [works](https://github.com/DataDog/datadog-agent-buildimages/actions/runs/13544912012/job/37854073536) so LGTM !

PR #772, opened 2025-02-26
Fix the temporarily existent tooling env var file

ofek:
### Motivation

https://github.com/DataDog/datadog-agent-macos-build/actions/runs/13553010764/job/37880903753

That repository uses the latest version of this repo and the file `deva.env` which was renamed a few hours ago exists only to prevent the builds over there from failing... the issue was that during the rename I was too eager and in fact this compatibility file must use the old environment variable names as well unlike what I thought


dev-envs/linux/utils.sh
```diff
@@ -143,9 +143,9 @@ GFOLD_VERSION="4.5.0"
 #     curl "https://github.com/nickgerace/gfold/releases/download/${GFOLD_VERSION}/gfold-linux-gnu-${short_arch}" -Lo /usr/local/bin/gfold
 #     chmod +x /usr/local/bin/gfold
 # else
-#   cargo install gfold@${GFOLD_VERSION}
+#   cargo install --locked gfold@${GFOLD_VERSION}
 # fi
-cargo install gfold@${GFOLD_VERSION}
+cargo install --locked gfold@${GFOLD_VERSION}
```
ofek, line 148: https://gitlab.ddbuild.io/DataDog/datadog-agent-buildimages/-/jobs/825114490#L4616

```
#18 9.863 error: failed to compile `gfold v4.5.0`, intermediate artifacts can be found at `/tmp/cargo-installAnjOp3`.
#18 9.863 To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.
#18 9.863 
#18 9.863 Caused by:
#18 9.863   package `zerofrom v0.1.6` cannot be built because it requires rustc 1.81 or newer, while the currently active rustc version is 1.76.0
#18 9.863   Try re-running cargo install with `--locked`
```

Same concept as https://github.com/DataDog/datadog-agent-buildimages/pull/768

PR #778, opened 2025-03-10
Update Linux developer image

ofek:
### Motivation

Use the build images with newer glibc everywhere


dev-envs/linux/default-vscode-extensions.txt
```diff
@@ -13,6 +13,5 @@ redhat.vscode-xml
 redhat.vscode-yaml
 rust-lang.rust-analyzer
 timonwong.shellcheck
-twxs.cmake
```
ofek, line 16: In the past few months this became deprecated and is now part of the main cmake extension

PR #779, opened 2025-03-12
glibc images: cmake: default to release builds

chouquette:
### What does this PR do?

Switch the default cmake build types to release

### Motivation

If we produce artifacts that aren't stripped by omnibus, we don't want to include the debug symbols


ofek (approved):
🥇

PR #782, opened 2025-03-13
fix Go package binaries by setting GOPATH

brycekahle:
### What does this PR do?

fix Go package binaries by setting GOPATH. I believe not setting `GOPATH` and adding it to the path was causing the below error.

### Motivation

Errors such as `/bin/bash: line 1: golangci-lint: command not found` from `dda inv -e install-tools`

### DIFF
--- a/linux-glibc-2.17-x64/Dockerfile
+++ b/linux-glibc-2.17-x64/Dockerfile
@@ -97,8 +97,10 @@ RUN /bin/bash -l -c "gem install bundler --version $BUNDLER_VERSION --no-documen
 RUN echo 'source /usr/local/rvm/scripts/rvm' >> /root/.bashrc
 
 # Go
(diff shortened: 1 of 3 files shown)


linux-glibc-2.17-x64/Dockerfile
```diff
@@ -97,8 +97,10 @@ RUN /bin/bash -l -c "gem install bundler --version $BUNDLER_VERSION --no-documen
 RUN echo 'source /usr/local/rvm/scripts/rvm' >> /root/.bashrc
 
 # Go
+ENV GOPATH=/go
 COPY setup_go.sh /
 RUN ./setup_go.sh
+ENV PATH="${GOPATH}/bin:${PATH}"
```
ofek, line 103: 🥇

ofek (approved):

PR #801, opened 2025-04-01
Persist the telemetry API key throughout container lifecycle

ofek:
### Motivation

Avoid having to pass the key during other user actions


ofek:
The build images are unaffected, only the Linux developer image, so I'm going to merge since everyone is offline

PR #803, opened 2025-04-02
Migrate invoke tasks to `dda` local commands

ofek:
This is basically a straight port except for upgraded type hints and using `httpx` over `requests`.

old | new
--- | ---
`dda inv agent.version` | `dda run agent version`
`dda inv update-datadog-agent-buildimages` | `dda run update agent-build-images`
`dda inv update-go` | `dda run update go`

Also, there is a new `dda run build` command which forwards all arguments to `docker build` and injects the proper `--build-arg` based on `*.env` files at the root of the repo. This currently is unused in CI because `dda` is not installed on hosts and increasing CI time even slightly might not be worth it although I'm interested in feedback about that.


CelianR (approved):
Nice!

misteriaud (changes requested):
Could you update `.github/workflows/go-update.yml` which is still referencing the invoke task and also update `.github/CODEOWNERS` so we are owners of the new `.dda/scripts/run/go` directory please

misteriaud (approved):
LGTM for ARUN code 🎉

PR #808, opened 2025-04-17
Do not build developer images on scheduled pipelines

ofek:
### Motivation

Build images do not get pushed on scheduled pipelines so there is no base image

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -193,7 +193,6 @@ lint_dockerfiles:
 .build_dev_env:
   stage: build
   rules:
-    - if: $CI_COMMIT_TAG == null
     - if: $CI_PIPELINE_SOURCE == "schedule"
       when: never
   script:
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -192,6 +192,10 @@ lint_dockerfiles:
 
 .build_dev_env:
   stage: build
+  rules:
+    - if: $CI_COMMIT_TAG == null
```
chouetz, line 196:
```suggestion
```
I think this is useless, because it has an implicit `when: on success` and we don't have tag most of the time (I'm not even sure we trigger pipelines on tags on this repo)

chouetz (approved):
Minor comment, otherwise LGTM

.gitlab-ci.yml
ofek: Should we remove this elsewhere (in a separate PR)?

https://github.com/DataDog/datadog-agent-buildimages/blob/6a9728054d16340b0523c1060c5e2a32b9e9c98c/.gitlab-ci.yml#L113-L116

chouetz: Indeed I think it has no effect so we can, thanks

PR #816, opened 2025-04-30
ci(refactor): Remove duplicate registry jobs

chouetz:
### What does this PR do?
- Remove the `registry` jobs as now we are using the same registry everywhere
- Use `parallel/matrix` when possible
- Remove hidden job when they were used only once

### Motivation
- Simplify our configuration
- Prepare migration to public-images for releasing

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -118,8 +118,6 @@ lint_dockerfiles:
     # Dockerhub login
     - DOCKER_REGISTRY_LOGIN=$(aws ssm get-parameter --region us-east-1 --name ci.datadog-agent.docker_hub_login --with-decryption --query "Parameter.Value" --out text)
     - aws ssm get-parameter --region us-east-1 --name ci.datadog-agent.docker_hub_pwd --with-decryption --query "Parameter.Value" --out text | docker login --username "$DOCKER_REGISTRY_LOGIN" --password-stdin docker.io
-    # ECR login
-    - eval "$(aws ecr get-login --region us-east-1 --no-include-email --registry-ids 486234852809)"
     # Build
     - GO_BUILD_ARGS=$(cat go.env | sed -e 's/^/--build-arg /' | tr '\n' ' ')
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -227,125 +185,79 @@ build_ci_image:
...
+      - CUSTOM_BUILD_ARGS: --build-arg KERNEL_EXTRA_CONFIG_VERSION=$EXTRA_KCONFIG_VERSION
+        DOCKERFILE: kernel-version-testing/kernel-version-testing_arm64/Dockerfile
+        IMAGE: kernel-version-testing_arm64
+      - DOCKERFILE: linux-glibc-2.23-arm64/Dockerfile
+        IMAGE: linux-glibc-2-23-arm64
 
 build_dev_env_linux_x64:
   extends: [.build_dev_env, .x64]
-  needs: [build_linux_glibc_2_17_x64]
+  # needs: [build_linux_glibc_2_17_x64]
```
CelianR, line 243: Should this be removed? Same for the other commented lines

.gitlab-ci.yml
```diff
@@ -408,11 +320,20 @@ push_to_datadog_agent:
...
   stage: build
   rules:
     - if: $CI_COMMIT_TAG == null
   timeout: 2h 00m
+  tags: ["windows-v2:2022"]
+  id_tokens:
+    CI_IDENTITY_GITLAB_JWT:
+      aud: https://vault.us1.ddbuild.io
+  variables:
+    CI_IDENTITY_ROLE_NAME_OVERRIDE: windows-ci-tmp-gitlab-id-token-datadog-agent-buildimages-all-refs
```
CelianR, line 333: Is it supposed to be this role? Since there is `tmp` in it

CelianR (approved):

.gitlab-ci.yml
chouetz: yes it's the role we should use.

.gitlab-ci.yml
chouetz: I commented for test but we must set a dependency, based on the new parallel:matrix reference

PR #845, opened 2025-05-20
Clean up Linux developer image

ofek:
### What does this PR do?

- Use the proper subprocess utility for newer versions of `dda` when running the `dda run build` local command
- Properly persist the `dda` environment variable to allow dynamic dependencies
- Remove duplicate Git installation now that the newer base image uses a modern version
- Enable high fidelity color output explicitly by default since the container cannot detect the terminal from which it runs

### DIFF
--- a/dev-envs/linux/dda.sh
+++ b/dev-envs/linux/dda.sh
@@ -2,8 +2,8 @@
 IFS=$'\n\t'
 set -euxo pipefail
 
-# Re-install dda if DDA_VERSION is set, otherwise link the dda from the conda environment
-if [[ -n "${DDA_VERSION}" ]]; then
+# Re-install dda if INSTALL_DDA is set, otherwise link the dda from the conda environment
+if [[ -n "${INSTALL_DDA:-}" ]]; then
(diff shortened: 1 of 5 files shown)


dev-envs/linux/dda.sh
```diff
@@ -2,8 +2,8 @@
 IFS=$'\n\t'
 set -euxo pipefail
 
-# Re-install dda if DDA_VERSION is set, otherwise link the dda from the conda environment
-if [[ -n "${DDA_VERSION}" ]]; then
+# Re-install dda if INSTALL_DDA is set, otherwise link the dda from the conda environment
+if [[ -n "${INSTALL_DDA:-}" ]]; then
```
chouetz, line 6:
```suggestion
if [[ -n "${INSTALL_DDA}" ]]; then
```
the `{VAR:-REPL}` will get REPL if VAR is not set, so I think it's useless here.

chouetz (approved):

dev-envs/linux/dda.sh
ofek: It's basically a way to check that it exists and it's not an empty string.

PR #850, opened 2025-05-27
glibc images: don't set DD_CC/DD_CXX

chouquette:
### What does this PR do?

Stop setting `DD_CC` & `DD_CXX` environment variable in the build images

### Motivation

Doing so causes build errors when building locally without omnibus, since the custom toolchains don't look for headers in the system locations.
Specifically, this prevents systemd headers from being found.

This is a work around until a more proper & long term solution is agreed upon


alopezz:
> @alopezz It will break the repackaging job, right? Any idea how we can keep it working?

https://github.com/DataDog/datadog-agent-buildimages/pull/849 was merged which for now fixes the original problem and for now it should work in combination with https://github.com/DataDog/datadog-agent/pull/37384. We're still planning to work on a better solution.

PR #860, opened 2025-06-04
feat(build): Cache the build results for linux jobs

chouetz:
### What does this PR do?
- Cache the build results using the `--cache-from` and `--cache-to` options from `docker build`. Cache in ECR to reuse the cache between builds
- Migrate the jobs to use k8s runners and buildkit
- Split the build jobs to use the various other docker commands (`docker inspect` or `docker history`) from the docker runner.
- Bump the images used on the runner to use more recent versions (get rid of ubuntu20)
- Split the configuration on several files as it started to become complex.

### Motivation
acix-878
contsec-1634

- No cache is possible now for windows: caching on ECR is only possible with buildkit and its [experimental](https://github.com/moby/buildkit/blob/master/docs/windows.md#experimental-windows-containers-support) on windows
- The PR is complex, sorry in advance for the reviewers.

### Additional Notes
This is a first step in improving buildimages. Next steps:
- multi stage image to benefit more from docker layering
- use build cache in windows or do a dynamic build


CelianR (approved):
Great!

PR #866, opened 2025-06-10
Improve cache handling

ofek:
### What does this PR do?

This removes more build artifacts and sets up the proper location for caching Ruby gems at runtime

### DIFF
--- a/dev-envs/linux/Dockerfile
+++ b/dev-envs/linux/Dockerfile
@@ -76,12 +76,9 @@ RUN /setup/vscode.sh
 COPY ssh.sh /setup/ssh.sh
 RUN /setup/ssh.sh
 
-# Remove setup files
-# hadolint ignore=DL3059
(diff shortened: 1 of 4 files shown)


Ishirui (approved):
One small concern: the `finalize.sh` script would probably be kind of annoying if it were to be run on a host machine by accident (maybe somebody fat-fingered their keyboard).

If that's something we're concerned about, would there be a way to add a check to make sure the script is being run in the context of a Dockerfile build ? (Maybe setting an env var or checking for some docker-specific file)

ofek:
None of the files are intended to be used in any other context other than building the image in that directory (there are a bunch of other scripts there as well).

ofek:
As another example, pretty much all scripts here either would not work locally or be undesirable: https://github.com/DataDog/datadog-agent-buildimages/tree/main/windows/helpers

PR #870, opened 2025-06-12
feat(windows): Enable buildkit and ECR cache for windows image

chouetz:
### What does this PR do?
- Enable [buildkit](https://github.com/moby/buildkit/blob/master/docs/windows.md) on the runner.
- Build the image using buildkit.
- Benefit from the ECR cache to speed-up the image build.

### Motivation
https://datadoghq.atlassian.net/browse/ACIX-878
Follow-up of https://github.com/DataDog/datadog-agent-buildimages/pull/860

- I removed the `docker rmi` which was done for each builds. I think this was impacting the local cache we expect on runners. I started a discussion with #ci-infra to check if we need to do this or if this is handled on their end.

### Additional Notes
- Job duration possible improvement:
image

- Discussed with CI/CD, we might want to create a fleet of runners with Buildkit enabled by default. It would save us from the runtime installation

### DIFF
--- a/buildkit.ps1
+++ b/buildkit.ps1
@@ -1,8 +1,11 @@
 # Source: https://github.com/moby/buildkit/blob/master/docs/windows.md#setup-instructions
-$url = "https://api.github.com/repos/moby/buildkit/releases/latest"
-$version = (Invoke-RestMethod -Uri $url -UseBasicParsing).tag_name
+. .\windows\helpers.ps1
+$version = "v0.22.0"
 $arch = "amd64" # arm64 binary available too
-curl.exe -fSLO https://github.com/moby/buildkit/releases/download/$version/buildkit-$version.windows-$arch.tar.gz
+$sha256 = "e76584227535814b25be9bb202ef400fd43c32b74c3ed3a0a05e9ceee40b3f66"
+$buildkit_url = "https://github.com/moby/buildkit/releases/download/$version/buildkit-$version.windows-$arch.tar.gz"
+$out = "$($PSScriptRoot)\buildkit-$version.windows-$arch.tar.gz"
+Get-RemoteFile -RemoteFile $buildkit_url -LocalFile $out -VerifyHash $sha256
 # there could be another `.\bin` directory from containerd instructions
 # you can move those
 mv bin bin2
--- a/containerd.ps1
+++ b/containerd.ps1
@@ -1,12 +1,19 @@
 # Source: https://github.com/containerd/containerd/blob/main/docs/getting-started.md#installing-containerd-on-windows
+. .\windows\helpers.ps1
 # If containerd previously installed run:
-Stop-Service containerd
+$service = Get-Service -Name containerd -ErrorAction SilentlyContinue
+if ($service) {
+    Stop-Service containerd
+}
 
 # Download and extract desired containerd Windows binaries
 $Version="1.7.13"	# update to your preferred version
 $Arch = "amd64"	# arm64 also available
-curl.exe -LO https://github.com/containerd/containerd/releases/download/v$Version/containerd-$Version-windows-$Arch.tar.gz
-tar.exe xvf .\containerd-$Version-windows-$Arch.tar.gz
+$sha256 = "a576160771eba9b3e0d85a841fa4f6dba60a14c5c4f45e34f2148e2fe138ebb7"
+$containerd_url = "https://github.com/containerd/containerd/releases/download/v$Version/containerd-$Version-windows-$Arch.tar.gz"
+$out = "$($PSScriptRoot)\containerd-$Version-windows-$Arch.tar.gz"
+Get-RemoteFile -RemoteFile $containerd_url -LocalFile $out -VerifyHash $sha256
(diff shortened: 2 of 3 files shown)


.gitlab/build.yml
```diff
@@ -157,12 +157,40 @@ build_windows_ltsc2022_x64:
...
-    - docker rmi $SRC_IMAGE
-    - If ($CI_JOB_STATUS -ne "success") { Write-Host "Build failed."; exit 0 }
-    - If ($CI_PIPELINE_SOURCE -ne "schedule") { Write-Host "Image $SRC_IMAGE is available." } else { Write-Host "Scheduled pipeline, image push skipped." }
+    - $CACHE_IMAGE = "registry.ddbuild.io/ci/datadog-agent-buildimages/${IMAGE}${ECR_TEST_ONLY}:cache"
+    # Install containerd, buildkit and CNI plugins, see https://github.com/moby/buildkit/blob/master/docs/windows.md
+    - .\containerd.ps1
+    - .\cni.ps1
+    - .\buildkit.ps1
+    # Start buildkitd
+    - Start-Process -FilePath "buildkitd.exe"
```
clarkb7, line 166: if it still makes sense, can you move this back into `build-container.ps1` ? It's helpful to be able to build the container locally, and having all the commands in the gitlab yaml makes that difficult to do. It's also difficult to read/maintain (no formatting/syntax highlighting)

.gitlab/build.yml
```diff
@@ -157,12 +157,40 @@ build_windows_ltsc2022_x64:
...
-  after_script:
-    - $SRC_IMAGE = "registry.ddbuild.io/ci/datadog-agent-buildimages/${IMAGE}${ECR_TEST_ONLY}:${IMAGE_VERSION}"
-    - docker rmi $SRC_IMAGE
-    - If ($CI_JOB_STATUS -ne "success") { Write-Host "Build failed."; exit 0 }
-    - If ($CI_PIPELINE_SOURCE -ne "schedule") { Write-Host "Image $SRC_IMAGE is available." } else { Write-Host "Scheduled pipeline, image push skipped." }
+    - $CACHE_IMAGE = "registry.ddbuild.io/ci/datadog-agent-buildimages/${IMAGE}${ECR_TEST_ONLY}:cache"
+    # Install containerd, buildkit and CNI plugins, see https://github.com/moby/buildkit/blob/master/docs/windows.md
+    - .\containerd.ps1
+    - .\cni.ps1
+    - .\buildkit.ps1
```
clarkb7, line 164: > Discussed with CI/CD, we might want to create a fleet of runners with Buildkit enabled by default. It would save us from the runtime installation

➕

buildkit.ps1
```diff
@@ -0,0 +1,21 @@
+# Source: https://github.com/moby/buildkit/blob/master/docs/windows.md#setup-instructions
+$url = "https://api.github.com/repos/moby/buildkit/releases/latest"
+$version = (Invoke-RestMethod -Uri $url -UseBasicParsing).tag_name
+$arch = "amd64" # arm64 binary available too
+curl.exe -fSLO https://github.com/moby/buildkit/releases/download/$version/buildkit-$version.windows-$arch.tar.gz
```
clarkb7, line 5: we should be pinning the version and validating the hash, for reproducibility as well as security reasons

clarkb7, line 5: might we run into rate limit issues? Maybe the volume of build-images jobs is low enough.

containerd.ps1
```diff
@@ -0,0 +1,29 @@
+# Source: https://github.com/containerd/containerd/blob/main/docs/getting-started.md#installing-containerd-on-windows
+# If containerd previously installed run:
+Stop-Service containerd
```
clarkb7, line 3: Do the runners run jobs in parallel? What happens if these installation scripts are run simultaneously on the same runner?

windows/Dockerfile
```diff
@@ -5,50 +5,18 @@ ARG BASE_IMAGE=mcr.microsoft.com/dotnet/framework/runtime:4.8-windowsservercore-
...
-LABEL go_version=${GO_VERSION}-${MSGO_PATCH}
-LABEL ruby_version=${RUBY_VERSION}
-LABEL wix_version=${WIX_VERSION}
-LABEL cmake_version=${CMAKE_VERSION}
-LABEL msys_version=${MSYS_VERSION}
-LABEL system_python_version=${PYTHON_VERSION}
-LABEL embedded_py3_version=${EMBEDDED_PYTHON_3_VERSION}
-LABEL nuget_version=${NUGET_VERSION}
-LABEL codeql_version=${CODEQL_VERSION}
-LABEL vault_version=${VAULT_VERSION}
```
clarkb7, line 51: Why are we removing these? idk what purpose they served originally tbh

buildkit.ps1
chouetz: You are right I picked up the code as is I can improve it a bit :)

chouetz: Yes I think the low rate is ok (and I hope this will be done on the runner once for all)

PR #874, opened 2025-06-16
Add Cursor to Linux dev env

ofek:
### What does this PR do?

1. Preinstalls Cursor server for remote SSH development just like we do for VS Code
2. Removes the C/C++ and Java extensions
    * `319.2M   ├─┴ms-vscode.cpptools-1.25.3-linux-x64`
    * `188.2M   ├─┴redhat.java-1.42.0-linux-x64`

### Motivation

Many of us now use Cursor, and few of us own Java/C/C++ code so there is no reason that their extensions should be installed by default

### Additional Notes

Although Cursor is a fork of VS Code, I was unable to use a shared directory for both because Cursor consistently runs behind on bumping the version they use which causes incompatibility amongst extensions:

- https://forum.cursor.com/t/vs-code-version/42129
- https://forum.cursor.com/t/outdated-vscode-extensions-may-contain-known-cves/50424

Therefore, each editor has their extensions linked to a subdirectory of the shared directory that is persisted as a volume.

### DIFF
--- a/dev-envs/linux/vscode.sh
+++ b/dev-envs/linux/vscode.sh
@@ -2,47 +2,122 @@
 IFS=$'\n\t'
 set -euxo pipefail
 
-VS_CODE_VERSION="1.98.0"
+VSCODE_VERSION="1.101.0"
+CURSOR_VERSION="1.1.3"
 
 arch=$(uname -m)
 if [[ $arch == "aarch64" ]]; then
-    DIGEST="1deb69c76288fb2f60a65fdcd0e5e4a484318cb822b88bc3f82ea51e36492460"
     arch="arm64"
+    VSCODE_SERVER_DIGEST="64a4f3030995ae2143d73edec32bf6213607125448157d7f3a485b2e7c547a91"
+    VSCODE_CLI_DIGEST="97b1f372294e8bd56a540d49d3224c92678a16cbd1d44178acf52f4ff2d81b36"
+    CURSOR_SERVER_DIGEST="d8d5cf4d9b73e144995d286ee91df0f5a3ada62f2517f76e1626d048cb8545b2"
+    CURSOR_CLI_DIGEST="5a2d4e75ec0f38aefc6d77968374f0ab20dd1f932175400a51ff04f29555fd62"
 else
-    DIGEST="7ac5717e59b09ac86b04786f6ff713e83d3744a0d936aeee0608917a765d986c"
     arch="x64"
-fi
-metadata_url="https://update.code.visualstudio.com/api/versions/${VS_CODE_VERSION}/linux-${arch}/stable"
-
-commit=$(curl -s "${metadata_url}" | jq -r .version)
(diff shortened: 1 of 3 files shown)


dev-envs/linux/vscode.sh
```diff
@@ -2,47 +2,122 @@
...
+    # We need to capture the first word of the name (e.g. `code` or `cursor`) for use later
+    bin_dir="${unpack_dir}/bin"
+    binary_name=""
+    pushd "${bin_dir}"
+    for binary in *; do
+        if [[ "${binary}" =~ ^([^-]+)-server$ ]]; then
+            binary_name="${BASH_REMATCH[1]}"
+        fi
+    done
+    popd
```
Ishirui, line 82: nit: I wasn't familiar with `pushd` and `popd`, might be clearer to just use `cd $OLDPWD` ?

```suggestion
    cd "${bin_dir}"
    for binary in *; do
        if [[ "${binary}" =~ ^([^-]+)-server$ ]]; then
            binary_name="${BASH_REMATCH[1]}"
        fi
    done
    cd "${OLDPWD}"
```

dev-envs/linux/vscode.sh
```diff
@@ -2,47 +2,122 @@
...
+cursor_server_url="https://cursor.blob.core.windows.net/remote-releases/${CURSOR_VERSION}-${cursor_commit_id}/vscode-reh-linux-${arch}.tar.gz"
+# Path from intercepting `wget` calls
+cursor_cli_url="https://cursor.blob.core.windows.net/remote-releases/${cursor_commit_id}/cli-alpine-${arch}.tar.gz"
+
+# Create shared extensions directory
+extensions_root_dir="${HOME}/.vscode-extensions"
+mkdir -p "${extensions_root_dir}"
+
+# Create function to install VS Code server
+function install_vscode() {
```
Ishirui, line 47: nit: Maybe we should call this `install_editor` or similar, since this is not only for vscode ? (although I guess it only works for vscode forks)

dev-envs/linux/vscode.sh
```diff
@@ -2,47 +2,122 @@
...
+    server_download_url="$3"
+    server_digest="$4"
+    cli_download_url="$5"
+    cli_digest="$6"
+
+    root_dir="${HOME}/${root_dir_name}"
+    install_dir="${root_dir}/cli/servers/Stable-${commit}"
+    unpack_dir="${install_dir}/server"
+
+    curl "${server_download_url}" -Lo "vscode_server.tar.gz"
```
Ishirui, line 59: nit: same as above, feels weird to be manipulating archives named "vscode" even when they might contain cursor files. Maybe we should use something more generic, like `editor_server.tar.gz` ?

Ishirui (approved):
Just a few nits :) Unfortunate that we can't share extensions directories :/

dev-envs/linux/vscode.sh
ofek: Yes I named it this because the file is only about installing VS Code and forks of it. I will update the comment.

PR #877, opened 2025-06-18
feat(linux): Remove ibm_mq from linux image

chouetz:
### What does this PR do?
Remove ibm_mq from linux image

### Motivation
There is no occurrence of running this tool in the `datadog-agent` repository so I suppose we can get rid of it.

### Additional Notes
[Full pipeline passed](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/pipelines/68634382) on `datadog-agent` side.


KevinFairise2:
Can we make sure that we are still able to build the integrations with that? That PR seems to indicate that it was removed from `deb_x64` image because it is no longer used to build the integrations, but I think the glibc image is now used to build the integrations in omnibus.build.
https://github.com/DataDog/datadog-agent-buildimages/pull/581

PR #900, opened 2025-07-10
Do not download artifacts at the beginning of the push-to-datadog-agent job

pgimalac:
### What does this PR do?
Update the push_to_datadog_agent job to avoid downloading artifacts from previous jobs.

### Motivation
It takes 20min to download artifacts, but they're not used.

### Additional Notes
I checked that it did remove downloading artifacts, and that the job still runs after previous stages.
https://gitlab.ddbuild.io/DataDog/datadog-agent-buildimages/-/jobs/1023098662


.gitlab/test.yml
```diff
@@ -54,6 +54,7 @@ push_to_datadog_agent:
     - !reference [.on_push]
   tags: ["arch:amd64"]
   image: registry.ddbuild.io/ci/datadog-agent-buildimages/deb_x64:v58799063-651eeeea
+  dependencies: [] # do not download artifacts from other jobs
```
chouetz, line 57: Sorry this is a consequence of using the k8s runners to build. I am not sure we really need the `stats` stage so it might not be necessary if we decide to get rid of it.

PR #793, opened 2025-03-27
Include cross compilers to the linux-glibc images

chouquette:
### What does this PR do?

Allow building for x86_64 & aarch64 from x86_64 and aarch64

### Motivation

This simplifies the life of the system-probe/EBPF teams

PR #902, opened 2025-07-11
feat(glibc): Use the same dockerfile

chouetz:
### What does this PR do?
Build the 2 glibc images (arm and amd) using the same dockerfile

### Motivation
- Prevent discrepancy between the images
- Ease the migration towards a single image in `datadog-agent`

There is no more the glibc version in the name of the image. We can add it as a label if we consider this is necessary

### Additional Notes
Tested on `datadog-agent`:
- [code impact](https://github.com/DataDog/datadog-agent/compare/main...nschweitzer/buildimages)
- [pipeline](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/pipelines/70199471)


Ishirui (approved):

rdesgroppes (approved):
@Ishirui's comments are worth considering, otherwise looks good to me.

linux-glibc/2.17/build.env
```diff
@@ -0,0 +1,12 @@
+CROSS_ARCH=aarch64
```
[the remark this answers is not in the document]
alopezz: That was added [here](https://github.com/DataDog/datadog-agent-buildimages/pull/793), which gives some (albeit terse) amount of context.

We can't generally build the Agent as a whole via cross-compilation, but I assume that some of the system-probe related components might be and that there's a use case for those.

.gitlab/build.yml
```diff
@@ -84,8 +61,10 @@ build_x64:
         IMAGE: kernel-version-testing_x64
       - DOCKERFILE: agent-deploy/Dockerfile
         IMAGE: gitlab_agent_deploy
-      - DOCKERFILE: linux-glibc-2.17-x64/Dockerfile
-        IMAGE: linux-glibc-2-17-x64
+      - DOCKERFILE: linux-glibc/Dockerfile
+        IMAGE: linux-glibc-x64
```
alopezz, line 65: > There is no more the glibc version in the name of the image.

Why not 
```suggestion
        IMAGE: linux-glibc-2-17-x64
```
then?

Or otherwise, if we want to deemphasize the glibc aspect of it, we may as well drop `glibc` from the name and name it after its main expected use, something like:
```suggestion
        IMAGE: linux-agent-build-x64
```

Ishirui: Definitely agree with the second suggestion, especially since the plan we discussed with @chouetz is to further merge all buildimages used in CI into one, i.e. "the one image ~~to rule them all~~ to build the whole agent". I don't see why we would emphasize the glibc aspect of it.

PR #929, opened 2025-08-01
Install authanywhere

CelianR:
### What does this PR do?

This is used to generate tokens to access internal services such as the stack cleaner, it's better to do it in the Dockerfile and not at runtime like it is done currently.

### DIFF
--- a/deb-arm/Dockerfile
+++ b/deb-arm/Dockerfile
@@ -258,6 +258,10 @@ RUN curl -LO https://releases.hashicorp.com/vault/${VAULT_VERSION}/${VAULT_FILEN
 RUN apt-get update \
     && apt-get install -y shellcheck
 
+# Install authanywhere for infra token management
+RUN curl -OL "binaries.ddbuild.io/dd-source/authanywhere/LATEST/authanywhere-linux-arm64" -o /bin/authanywhere && \
+    chmod +x /bin/authanywhere
+
 RUN git config --global user.email "package@datadoghq.com"
 RUN git config --global user.name "Bits"
 
--- a/deb-x64/Dockerfile
+++ b/deb-x64/Dockerfile
@@ -250,6 +250,10 @@ RUN curl -LO https://github.com/protocolbuffers/protobuf/releases/download/v${PR
 RUN apt-get update \
     && apt-get install -y shellcheck
 
+# Install authanywhere for infra token management
+RUN curl -OL "binaries.ddbuild.io/dd-source/authanywhere/LATEST/authanywhere-linux-amd64" -o /bin/authanywhere && \
+    chmod +x /bin/authanywhere
+
 # create the agent build folder within $GOPATH
 RUN mkdir -p /go/src/github.com/DataDog/datadog-agent
 


rdesgroppes (approved):
Better than fetching `LATEST` on the fly.

ofek:
> it should be only necessary on deb arm / x86

Why?

CelianR:
> > it should be only necessary on deb arm / x86
> 
> Why?

@ofek I think that our tasks relying on github / gitlab tokens are only using these images but tell me if you have an image in mind that uses these tokens != deb

ofek:
I think it would be better to install these not just where they are currently used but also on the glibc images since those are going to be used for everything soon.

ofek (changes requested):
Can we pin the version with a hash like the other tools?

ofek (approved):

setup/authanywhere.sh
```diff
@@ -0,0 +1,16 @@
...
+    authanywhere_arch="arm64"
+else
+    echo "Unsupported architecture to install authanywhere: $(uname -m)"
+    exit 1
+fi
+
+# Install authanywhere for infra token management
+curl -OL "binaries.ddbuild.io/dd-source/authanywhere/LATEST/authanywhere-linux-$authanywhere_arch" && \
+    mv "authanywhere-linux-$authanywhere_arch" /bin/authanywhere && \
+    chmod +x /bin/authanywhere
```
rdesgroppes, line 16:
```suggestion
curl -OL "binaries.ddbuild.io/dd-source/authanywhere/LATEST/authanywhere-linux-$authanywhere_arch"
mv "authanywhere-linux-$authanywhere_arch" /usr/local/bin/authanywhere
chmod +x /usr/local//bin/authanywhere
```
- no longer being in a Dockerfile, the chain of `&&` is no longer needed (and we have the fail-fast thanks to `-e`)
- prefer `/usr/local/bin` over `/bin`:
    - `/bin` is deprecated on modern Linux distros (either a symlink to `/usr/bin` or even absent)
    - for packages installed out of package manager control (FHS)
    - also because of precedence in "canonical" `PATH`s (usually `...:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:...`)

setup/authanywhere.sh
```diff
@@ -0,0 +1,18 @@
...
+    authanywhere_arch="amd64"
+elif [ "$(uname -m)" == "aarch64" ]; then
+    authanywhere_arch="arm64"
+else
+    echo "Unsupported architecture to install authanywhere: $(uname -m)"
+    exit 1
+fi
+
+# Install authanywhere for infra token management
+curl -OL https://binaries.ddbuild.io/dd-source/authanywhere/v0.0.2/authanywhere-tar.tar.gz
```
rdesgroppes, line 14: sha256sum --check --strict <(echo "05d14b25e4607cc9e14867f2a0f38774869ae609eec890168e57de9a1b428e37 *authanywhere-tar.tar.gz")

rdesgroppes:
Thanks for your persistence, @CelianR!

PR #930, opened 2025-08-01
Properly handle errors in Windows build scripts

ofek:
### Motivation

Most other scripts had that first line to catch all non-zero exit codes from cmdlets

https://gitlab.ddbuild.io/DataDog/datadog-agent-buildimages/-/jobs/1060604437

Ultimately same error as before but now errors are caught properly (for cmdlets) https://github.com/DataDog/datadog-agent-buildimages/pull/925#issuecomment-3139738873

### DIFF
--- a/windows/helpers/phase2/install_python.ps1
+++ b/windows/helpers/phase2/install_python.ps1
@@ -3,6 +3,7 @@ param (
     [Parameter(Mandatory=$true)][string]$Sha256
 )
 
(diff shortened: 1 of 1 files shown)


ofek:
Now Windows images can be properly released. The silent error meant that it never got pushed during the build https://gitlab.ddbuild.io/DataDog/public-images/-/jobs/1060733292

PR #939, opened 2025-08-07
Small improvements to Linux builds

ofek:
### What does this PR do?

Starts builds of the Linux developer environment immediately after the required base images are available rather than after everything.

### DIFF
--- a/.gitlab/build.yml
+++ b/.gitlab/build.yml
@@ -126,7 +126,7 @@ build_multiarch:
 build_dev_env_linux_x64:
   extends: [.build, .no_schedule, .x64]
   needs:
-    - job: build_multiarch
+    - job: build_x64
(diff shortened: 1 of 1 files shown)


Ishirui (changes requested):
Don't think that will work as is :confused:
Since the `build_dev_env` build jobs still use the normal `.build` template they'll attempt to pull the image with the usual `v{PIPELINE_ID}_{COMMIT_SHA}` tag, instead of the arch-specific `v{PIPELINE_ID}_{COMMIT_SHA}-{ARCH}` pushed by the `build_x64` and `build_arm64` jobs for the linux images.

We either need to override `$IMAGE_VERSION` in here too (which, tbh, is not that bad), or let them depend on the multiarch one.

PR #941, opened 2025-08-12
Remove legacy Debian images

ofek:
### Motivation

They are no longer used on `main` https://github.com/DataDog/datadog-agent/pull/39535

We retain the ability to build older releases because there are dedicated branches for each one.


rdesgroppes (approved):
LG

Issue #942, opened 2025-08-12
[BUG] warning prints triggered on every call to a dda command


lavigne958:
**Environment**
linux / macos

**Describe what happened:**
When using devcontainers, any time I call the command `dda` I receive a warning that is printed on the console due to some bug in a python library.

```
datadog@agent-devenv:~/go/src/github.com/DataDog/datadog-agent$ dda inv test --help

/usr/lib/python3/dist-packages/blinker/base.py:96: SyntaxWarning: invalid escape sequence '\*'
  `sender=` as a single positional argument and any \*\*kwargs that
/usr/lib/python3/dist-packages/blinker/base.py:174: SyntaxWarning: invalid escape sequence '\*'
  `sender=` as a single positional argument and any \*\*kwargs that
/usr/lib/python3/dist-packages/blinker/base.py:242: SyntaxWarning: invalid escape sequence '\*'
  """Emit this signal on behalf of *sender*, passing on \*\*kwargs.
Usage: inv[oke] [--core-opts] test [--options] [other tasks here ...]
```

**Describe what you expected:**

i expected the command to run without the warnings on python code, like:

```
datadog@agent-devenv:~/go/src/github.com/DataDog/datadog-agent$ dda inv test --help
Usage: inv[oke] [--core-opts] test [--options] [other tasks here ...]
```

**Steps to reproduce the issue:**

Use the latest know docker image for the devcontainer: `registry.ddbuild.io/ci/datadog-agent-devenv:1-arm64`

start a devcontainer using the compose.yaml file.

run the dda command: `dda`

**Additional environment details (Operating System, Cloud provider, etc):**

PR #943, opened 2025-08-12
Install latest version of blinker, upgrade dda version

lavigne958:
### What does this PR do?

Fixup warnings emitted on every call to the command `dda` by installing latest version of the python package `blinker` in usersapce.

Upgrade dda version.

### Motivation

It's very anoying to have warning printed and taking the output.

None that I see so far.

### Additional Notes

Closes: #942

### DIFF
--- a/devcontainer/Dockerfile
+++ b/devcontainer/Dockerfile
@@ -74,6 +74,10 @@ RUN useradd -g 20 -u 503 -m datadog -s /bin/bash -G sudo,root && \
     apt-get autoremove -y && apt-get clean && rm -rf /var/lib/apt/lists
 
 USER datadog
+
+# prevents using system package which emits warning logs on every call
+RUN pip3 install --no-cache-dir blinker==1.9.0
+
 ENV GOPATH=/home/datadog/go \
     GOROOT=/usr/local/go \
     PATH="/home/datadog/.local/bin:/home/datadog/go/bin:/usr/local/go/bin:${PATH}"
--- a/devcontainer/README.md
+++ b/devcontainer/README.md
@@ -29,8 +29,16 @@ docker pull registry.ddbuild.io/ci/datadog-agent-devenv:1-<platform>
 
 Option 2: Build the image locally
 
+Set the DDA version
+
 ```
-DOCKER_BUILDKIT=1 docker build --build-arg="GO_VERSION=1.22.5" -t registry.ddbuild.io/ci/datadog-agent-devenv:1 .
+source .dda.env
+```
+
+build the docker image
+
+```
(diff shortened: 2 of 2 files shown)


lavigne958:
> The version of `dda` is managed entirely by this file https://github.com/DataDog/datadog-agent-buildimages/blob/main/dda.env
> 
> Are you sure you want to document a hardcoded value elsewhere?

did not know about it, In this case I don't.
I could update the documentation to extract the version from the file then run the docker build command with the right value extracted from the file. 

something like this:

Option 2: Build the image locally

1.  get DDA version
```
<repository root dir> $ source dda.env
```

2. build the docker image
```
<repository root dir> $ docker build --build-arg="GO_VERSION=1.24.5" --build-arg=DDA_VERSION=${DDA_VERSION} -t registry.ddbuild.io/ci/datadog-agent-devenv:1 .
```

clamoriniere:
> Why not just bump the version if you require the new version? It's passed in automatically to the image build as an argument.

Hi @ofek, 
I don't think it is the case for the devcontainer image. I see the usage of `dda.env` in the [build.sh](https://github.com/DataDog/datadog-agent-buildimages/blob/main/build.sh) script and then in the [gitlab.yml build job](https://github.com/DataDog/datadog-agent-buildimages/blob/5bdfdb3ebc6342b0f6331bfb2a5e4e2326eaed8a/.gitlab/build.yml#L38). However the devcontainer image is not using this script and rely on another gilab job https://github.com/DataDog/datadog-agent-buildimages/blob/main/.gitlab/devcontainer.yml 
for now it seams harcoded in the gitlab job: https://github.com/DataDog/datadog-agent-buildimages/blob/5bdfdb3ebc6342b0f6331bfb2a5e4e2326eaed8a/.gitlab/devcontainer.yml#L22 

@lavigne958 the solution seems to revert this PR https://github.com/DataDog/datadog-agent-buildimages/pull/842 if now the problem has been fixed.

@ofek could you confirm we can now revert your previous [PR](https://github.com/DataDog/datadog-agent-buildimages/pull/842)? 🙇

ofek:
Ah I forgot about that, thanks! I don't think reverting will work as the underlying issue to my knowledge has not been fixed and our efforts have been geared toward making the new developer image satisfy the needs of everyone. Incidentally, I just opened a fix for the final issue that was blocking the containers team from adopting the new image https://github.com/DataDog/datadog-agent/pull/39793

I'm going to sync with Wassim after that is merged to see about what else (if anything) is blocking and discuss the strategy for communication of the switch.

lavigne958:
> As a follow-up to my previous comment, today I spoke with both @wdhif and @L3n41c regarding the switch to the new [development environments](https://datadoghq.dev/datadog-agent/tutorials/dev/env/). They now seem content with the capabilities of the new image but they are going to confirm by testing tomorrow or early next week.

Perfect, thank you for taking the time to check all this.
I'll wait for their response on that matter if positive then we can merge this.

PR #951, opened 2025-08-14
[ABLD-135] Add `bazelisk` to Linux containers

rdesgroppes:
### What does this PR do?
As the title implies, this is to install the `bazelisk` command line tool (Go binary, ~9 MiB before compression) meant to bootstrap `bazel`, since the version of `bazel` itself will be branch-pinned for reproducible builds.
    
The present iteration targets Linux containers, with the noticeable exception of the `armhf` variant (ARM hard-float, 32-bit architecture) for which no `bazel` distribution will ever exist. We therefore aim at addressing the need by using cross-compilation instead - through `bazel` on either `x86_64` (aka `amd64`) or `aarch64` (aka `arm64`).
    
Note: the change leverages heredocs in `Dockerfile`s, which came out[^1] in 2021 and is now fully supported by:
- BuildKit v0.10.0+, since August 2022,
- Docker 23.0+, since February 2023. (without a need for passing experimental flags)
    
### Motivation
See [Agent Bazel Migration Plan](https://datadoghq.atlassian.net/wiki/spaces/ABLD/pages/5371789343/Agent+Bazel+Migration+Plan).

We may later envision to switch to the standalone binary distribution of `bazelisk` - at the expense of an increased filesystem footprint though.

### Additional Notes
See also:
- DataDog/ci-platform-machine-images#395
[^1]: https://www.docker.com/blog/introduction-to-heredocs-in-dockerfiles/

### DIFF
--- a/agent-deploy/Dockerfile
+++ b/agent-deploy/Dockerfile
@@ -177,6 +177,16 @@ ENV GO_VERSION $GO_VERSION
 ENV PATH="$PATH:/usr/local/go/bin"
 ENV PATH="${GOPATH}/bin:${PATH}"
 
+# Install & verify Bazelisk (Go package) as Bazel bootstrapper
+RUN bash <<'EOF'
+    set -euxo pipefail
+    go install github.com/bazelbuild/bazelisk@v1.27.0
+    ln -s bazelisk "$(command -v bazelisk | sed 's/bazelisk$/bazel/')"
+    export BAZELISK_HOME="$(mktemp -d)"
+    trap 'rm -rv -- "$BAZELISK_HOME"' EXIT
+    USE_BAZEL_VERSION=7.6.1 bazel --version | grep -Fq 'bazel 7.6.1'
+EOF
+
 # Install vault: https://github.com/hashicorp/vault/blob/main/CHANGELOG.md https://releases.hashicorp.com/vault
 RUN curl -LO https://releases.hashicorp.com/vault/${VAULT_VERSION}/${VAULT_FILENAME} && \
   echo "${VAULT_CHECKSUM} ${VAULT_FILENAME}" | sha256sum --check && \
--- a/devcontainer/Dockerfile
+++ b/devcontainer/Dockerfile
@@ -84,5 +84,16 @@ RUN go install golang.org/x/tools/gopls@latest && \
     go install github.com/go-delve/delve/cmd/dlv@latest && \
     go install github.com/golangci/golangci-lint/cmd/golangci-lint@latest && \
     go install gotest.tools/gotestsum@latest
+
+# Install & verify Bazelisk (Go package) as Bazel bootstrapper
+RUN bash <<'EOF'
+    set -euxo pipefail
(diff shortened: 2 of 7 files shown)


agent-deploy/Dockerfile
```diff
@@ -177,6 +177,16 @@ ENV GO_VERSION $GO_VERSION
...
 ENV PATH="${GOPATH}/bin:${PATH}"
 
+# Install & verify Bazelisk (Go package) as Bazel bootstrapper
+RUN bash <<'EOF'
+    set -euxo pipefail
+    go install github.com/bazelbuild/bazelisk@v1.27.0
+    ln -s bazelisk "$(command -v bazelisk | sed 's/bazelisk$/bazel/')"
+    export BAZELISK_HOME="$(mktemp -d)"
+    trap 'rm -rv -- "$BAZELISK_HOME"' EXIT
+    USE_BAZEL_VERSION=7.6.1 bazel --version | grep -Fq 'bazel 7.6.1'
```
JSGette, line 187: #ff2r 
Why using this bazel version and not something more recent?

devcontainer/Dockerfile
```diff
@@ -84,5 +84,16 @@ RUN go install golang.org/x/tools/gopls@latest && \
     go install github.com/go-delve/delve/cmd/dlv@latest && \
     go install github.com/golangci/golangci-lint/cmd/golangci-lint@latest && \
     go install gotest.tools/gotestsum@latest
+
```
JSGette, line 87: Just out of curiousity: don't we have a habit of just having a shell script? Since the block is the same across several  different Dockerfiles I would rather reuse a shell script and do something like:
`RUN --mount=type=bind,src=install_bazelisk.sh,target=/tmp/install_bazelisk.sh \
    chmod +x /tmp/install_bazelisk.sh && /tmp/install_bazelisk.sh`

This way you don't need to `COPY` that script it into the layer

rdesgroppes: I like it... but will leave it to a subsequent PR dedicated to that refactoring, if that's fine for you. (I don't pretend I'll do it right after, nor whether I'll be the one doing it :wink:)

agent-deploy/Dockerfile
rdesgroppes: tl;dr: it really _should_ not matter: it's just a arbitrarily picked-by-me older version, hardcoded for verification purposes.

Now, the rationale...
Why 7.6.1 and not 8.3.1, for instance? Well, in an earlier life, I had to maintain a `Dockerfile` based on an upstream image. Things looked fine, but [POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) got defeated: someone had the bright idea to preinstall a recent version of `bazel` in the upstream image (whose path was taking precedence), which we realized way later when attempting to bump our main branch's `.bazelversion`: effective for developers but... ineffective in CI! (that was using the derived image) Not to mention the extra work incurred by looking for the culprit and retrofitting our long-term maintenance branches. That's why I'm now used to verify `bazelisk` installs a not-to-recent version of `bazel`...

PR #967, opened 2025-08-20
ci(stats): Compute stats using crane instead of docker commands

chouetz:
### What does this PR do?
Compute image stats using crane instead of docker

### Motivation
Get rid of the `stat` stage which used docker runners and need artefact retention

The figures are totally different (and much lower) but the order and ratio between images [seem to be similar](https://app.datadoghq.com/metric/explorer?fromUser=false&state=H4sIAAAAAAAAE12R3WrDMAyF30WXw5QkXbo1rzKKMbGcCvyTWk4hDXn3oaTtYHfS4fPRkbzAHTNTitBBUzWfuq51VYOCIZvxytD9LGDRUaSyQQsUKh6hA1B7qZke0tent2I8DWLo0RUR51GAQgEZMyGDgoy3CbnsAzLymCKjdikHU_6zt2mvBI0miJdIs8S0phjNacq9yAFLpv71ZoYOzH3oeOaC4dCP02FizMvHCutFgQybvNl9n82f9XpZL6tgPHoq2lLAKHcS_K32KToaNgNPgQp0daWAUy5yqZQtZujAIveweclaLm8rLMDFCFd_te2pac7f52PbKsBod-2rOh-fWkaXka86JCtbsidLcQAFo5kYLXTOeMZVgSNfMD8jviPrOz306xPS6IkLrL_qvWqQ-QEAAA&start=1755622989355&end=1755709389355&paused=false#N4Ig7glgJg5gpgFxALlAGwIYE8D2BXJVEADxQEYAaELcqyKBAC1pEbghkcLIF8qo4AMwgA7CAgg4RKUAiwAHOChASAtnADOcAE4RNIKtrgBHPJoQaUAbVBGN8qVoD6gnNtUZCKiOq279VKY6epbINiAiGOrKQdpYZAYgUJ4YThr42gDGSsgg6gi6mZaBZnHKGABuMMjJCBhQODAAdABGeBBoUD4Y8BpNGhAAXnDALdoYIpmMyCIaU2DsCMPaAPQadRY8AAQtWFvA3fA8IHygkdG5sVgATIm1qel4WTl5iIXFIFflVTUpDc1tDpdDy9fpDEZjCZTZAeUTbXb7Q5wY48AC6VFc7jwmFC4UxqmxGBipXiJwooHWWDQOVA8gw1IQCBemRwODQZJA+MJxJ0NxO6JAlOpMhAdIZTOU-ycmRxA0yiTQojgTjkimU6UVUAVSqc9CYyhEbg87L4SQg9kwWBVCheipESjRPFNFvEAGEpMIYCgRNi0DwgA). As we are mostly concerned by the evolution rather than the raw value I think we can do the switch.


ofek (approved):
Do you have a sense of which measurement is more accurate?

chouetz:
> Do you have a sense of which measurement is more accurate?
@ofek 

docker inspect -f "{{ .Size }}" theImage

This command returns the uncompressed size of the image as stored locally in Docker's storage. This includes:
•  The total size of all layers when they're extracted/uncompressed
•  Any overhead from Docker's storage driver
•  The actual disk space the image occupies on your local system

crane manifest theImage | jq '.config.size + ([.layers[].size] | add)'

This command calculates the compressed size by examining the image manifest. It adds:
•  The config blob size (.config.size)
•  The sum of all compressed layer sizes ([.layers[].size] | add)

This represents the size as stored in the registry and what would be transferred during a docker pull.

I think we can rely on the latter one, and if we need to do precise measurements we can do them locally using the docker command.

ofek:
Gotcha, thank you!

PR #993, opened 2025-09-26
[ACIX-1069] Fix docker image caching

Ishirui:
### What does this PR do?

Fixes the Docker image caching logic that was broken after we merged the two linux images into a multiarch image.

Also introduces a new caching strategy: each build pull the cache from its own "branch-specific" cache image as well as `main`, but only pushes to its own branch-specific cache image.

This way the `main` cache only gets updated when a PR is merged, but PR builds can benefit from the cached layers coming from builds on `main`.

### Motivation

Increasing the speed of CI, as image builds can sometimes take quite a while (>1h for the main `linux` image)
This should also help with the `buildx`-related hangs we've been seeing.

Increased storage use on ECR - need to figure out a cleanup policy.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -7,10 +7,9 @@ if [[ "$DOCKERFILE" == "dev-envs/linux/Dockerfile" ]]; then WORKDIR="dev-envs/li
 
 # == Caching logic == #
 function sanitize() {
-    # From docker docs: https://docker-docs.uclv.cu/engine/reference/commandline/tag/#extended-description
-    # A tag name must be valid ASCII and may contain lowercase and uppercase letters, digits, underscores, periods and dashes.
+    # Docker uses the Go reference format for images: https://pkg.go.dev/github.com/distribution/reference#pkg-overview
+    # A tag name must match the following regex: /[\w][\w.-]{0,127}/
     # Git branch names can contain disallowed characters, so we need to sanitize them.
-    # xargs is used to remove leading and trailing whitespace.
     echo "$1" | tr -C '\na-zA-Z0-9_.-' '_'
 }
 
--- a/build.sh
+++ b/build.sh
@@ -4,13 +4,14 @@ set -euo pipefail
 # Build and push to internal ECR
 WORKDIR="."
 if [[ "$DOCKERFILE" == "dev-envs/linux/Dockerfile" ]]; then WORKDIR="dev-envs/linux"; fi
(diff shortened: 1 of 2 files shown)


Ishirui:
As we can see by the times and logs on the previous two commit's builds, this works _better_: we get some `CACHED` layers, and the time is sometimes completely down ([example](https://gitlab.ddbuild.io/DataDog/datadog-agent-buildimages/-/jobs/1150015775))

However, for some reason sometimes the layers still get rebuilt even though they should not / should be present in the cache. ([example](https://gitlab.ddbuild.io/DataDog/datadog-agent-buildimages/-/jobs/1150015792), from the same pipeline as above). Not sure _why_ this happens, but I suspect this is due to the slew of arguments getting interpolated in the command itself, as that should be the only comparison key used for `RUN` commands' cache hits.

build.sh
```diff
@@ -1,16 +1,42 @@
...
+set -euxo pipefail
 
 # Build and push to internal ECR
 WORKDIR="."
 if [[ "$DOCKERFILE" == "dev-envs/linux/Dockerfile" ]]; then WORKDIR="dev-envs/linux"; fi
-CACHE_SOURCE="--cache-from type=registry,ref=registry.ddbuild.io/ci/datadog-agent-buildimages/$IMAGE${ECR_TEST_ONLY}:cache"
+
+# == Caching logic == #
+function sanitize() {
+    # From docker docs: https://docker-docs.uclv.cu/engine/reference/commandline/tag/#extended-description
```
ofek, line 10: We should use the official documentation, I'm not sure what this is.

ofek (changes requested):

build.sh
```diff
@@ -1,16 +1,41 @@
...
 WORKDIR="."
 if [[ "$DOCKERFILE" == "dev-envs/linux/Dockerfile" ]]; then WORKDIR="dev-envs/linux"; fi
-CACHE_SOURCE="--cache-from type=registry,ref=registry.ddbuild.io/ci/datadog-agent-buildimages/$IMAGE${ECR_TEST_ONLY}:cache"
+
+# == Caching logic == #
+function sanitize() {
+    # Docker uses the Go reference format for images: https://pkg.go.dev/github.com/distribution/reference#pkg-overview
+    # A tag name must match the following regex: /[\w][\w.-]{0,127}/
+    # Git branch names can contain disallowed characters, so we need to sanitize them.
+    echo "$1" | tr -C '\na-zA-Z0-9_.-' '_'
```
CelianR, line 13: I'm not sure to understand, is `\n` there to pass multiple lines or should we remove it and sanitize it?

CelianR (approved):
LGTM, you can wait on Ofek's review as well

PR #1129, opened 2026-04-09
Add Renovate auto-update for datadog-ci

chouetz:
## Summary
- Add a Renovate regex manager to track `DATADOG_CI_VERSION` in `docker-bake.hcl` against `DataDog/datadog-ci` GitHub releases
- Add a GitHub Actions workflow (`.github/workflows/update-datadog-ci.yml`) that triggers on Renovate pushes to fetch checksums and propagate version + SHA updates to all Dockerfiles and `windows/versions.ps1`

## How it works
1. Renovate detects a new `DataDog/datadog-ci` release (with 14-day cooldown) and bumps the version in `docker-bake.hcl`
2. The push to the `renovate/**` branch triggers the workflow
3. The workflow downloads `checksums.txt` from the release, then updates all files with the correct per-architecture SHA256 and version
4. Changes are committed and pushed back to the Renovate PR branch

## Test plan
- [ ] Verify Renovate picks up the new regex manager (check Renovate dashboard)
- [ ] Manually trigger the workflow on a test branch to validate the sed replacements
- [ ] Confirm the workflow correctly updates all 8 target files

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/update-datadog-ci.yml
+++ b/.github/workflows/update-datadog-ci.yml
@@ -49,14 +49,10 @@ jobs:
           sed -i '/^variable "checksums_arm64"/,/^}/{ /DATADOG_CI_SHA256/s/[a-f0-9]\{64\}/'"${SHA_LINUX_ARM64}"'/ }' docker-bake.hcl
 
           # Update version + x64 SHA in x64 Dockerfiles
-          for f in rpm-x64/Dockerfile docker-x64/Dockerfile dd-agent-testing/Dockerfile; do
+          for f in rpm-x64/Dockerfile docker-x64/Dockerfile dd-agent-testing/Dockerfile agent-deploy/Dockerfile; do
             sed -i "/^ARG DATADOG_CI_VERSION/s/=.*/=${VERSION}/" "$f"
             sed -i "/^ARG DATADOG_CI_SHA/s/[a-f0-9]\{64\}/${SHA_LINUX_X64}/" "$f"
           done
-
-          # agent-deploy uses DATADOG_CI_SHASUM instead of DATADOG_CI_SHA
-          sed -i "/^ARG DATADOG_CI_VERSION/s/=.*/=${VERSION}/" agent-deploy/Dockerfile
-          sed -i "/^ARG DATADOG_CI_SHASUM/s/[a-f0-9]\{64\}/${SHA_LINUX_X64}/" agent-deploy/Dockerfile
 
           # Update version + arm64 SHA in arm64 Dockerfiles
           for f in rpm-arm64/Dockerfile docker-arm64/Dockerfile; do
--- a/.github/workflows/update-datadog-ci.yml
+++ b/.github/workflows/update-datadog-ci.yml
@@ -1,0 +1,81 @@
+name: Update datadog-ci checksums
+
+on:
+  push:
+    branches:
+      - 'renovate/**'
+    paths:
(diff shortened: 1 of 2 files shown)


renovate.json
```diff
@@ -25,6 +25,17 @@
       ],
       "depNameTemplate": "datadog/datadog-agent-dev",
       "datasourceTemplate": "github-releases"
+    },
+    {
+      "customType": "regex",
+      "fileMatch": [
+        "docker-bake\\.hcl"
+      ],
```
Ishirui: Feels weird to only update the bakefile here, could we not do the same with the ARGs in the Dockerfiles ?

.github/workflows/update-datadog-ci.yml
```diff
@@ -0,0 +1,81 @@
...
+        id: version
+        run: |
+          VERSION=$(grep 'DATADOG_CI_VERSION' docker-bake.hcl | head -1 | grep -oE '[0-9]+\.[0-9]+\.[0-9]+')
+          echo "version=${VERSION}" >> "$GITHUB_OUTPUT"
+
+          curl -fsSL "https://github.com/DataDog/datadog-ci/releases/download/v${VERSION}/checksums.txt" -o /tmp/checksums.txt
+
+          SHA_LINUX_X64=$(awk '/datadog-ci_linux-x64$/{print $1}' /tmp/checksums.txt)
+          SHA_LINUX_ARM64=$(awk '/datadog-ci_linux-arm64$/{print $1}' /tmp/checksums.txt)
+          SHA_WIN_X64=$(awk '/datadog-ci_win-x64$/{print $1}' /tmp/checksums.txt)
```
Ishirui, line 34: We could probably get it directly from the GitHub API (you can see the SHA256 displayed in the UI too) instead of downloading a file, not sure that that is better though

renovate.json
Ishirui: Disregard - I see now that the action handles that

.github/workflows/update-datadog-ci.yml
```diff
@@ -0,0 +1,81 @@
...
+
+          # Update version + x64 SHA in x64 Dockerfiles
+          for f in rpm-x64/Dockerfile docker-x64/Dockerfile dd-agent-testing/Dockerfile; do
+            sed -i "/^ARG DATADOG_CI_VERSION/s/=.*/=${VERSION}/" "$f"
+            sed -i "/^ARG DATADOG_CI_SHA/s/[a-f0-9]\{64\}/${SHA_LINUX_X64}/" "$f"
+          done
+
+          # agent-deploy uses DATADOG_CI_SHASUM instead of DATADOG_CI_SHA
+          sed -i "/^ARG DATADOG_CI_VERSION/s/=.*/=${VERSION}/" agent-deploy/Dockerfile
+          sed -i "/^ARG DATADOG_CI_SHASUM/s/[a-f0-9]\{64\}/${SHA_LINUX_X64}/" agent-deploy/Dockerfile
```
Ishirui: nit: why not rename the variable in this PR ? It shouldn't be that complicated a change

Ishirui (approved):
LGTM, just a few comments and tweaks

.github/workflows/update-datadog-ci.yml
chouetz: I think fetching this file is the simplest approach, as you do a single call and retrieve all the information you need

PR #1141, opened 2026-04-13
Add mold linker to buildimages

pgimalac:
### What does this PR do?
Add mold linker to linux buildimages and devcontainer.

### Motivation
Speed up builds (both in CI and locally).

### DIFF
--- a/devcontainer/Dockerfile
+++ b/devcontainer/Dockerfile
@@ -9,6 +9,9 @@ ARG PYTHON_VERSION=3.12
 ARG CMAKE_VERSION=3.30.2
 ARG CMAKE_SHA256_AMD64="33f5a7680578481ce0403dc5a814afae613f2f6f88d632a3bda0f7ff5f4dedfc"
 ARG CMAKE_SHA256_ARM64="8a6636e72a6ddfe50e0087472bff688f337df48b00a7728b12d7b70b5b459fc5"
+ARG MOLD_VERSION=2.40.4
+ARG MOLD_SHA256_AMD64="4c999e19ffa31afa5aa429c679b665d5e2ca5a6b6832ad4b79668e8dcf3d8ec1"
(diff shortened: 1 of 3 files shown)


pgimalac:
@codex review

devcontainer/Dockerfile
```diff
@@ -49,6 +52,19 @@ RUN if [ "$TARGETARCH" = "amd64" ] ; then set -ex \
...
+    && echo "${MOLD_SHA256_ARM64}  /tmp/mold.tar.gz" | sha256sum --check \
+    && tar -C /usr/local --strip-components=1 -xzf /tmp/mold.tar.gz \
+    && rm /tmp/mold.tar.gz ; fi
+
+RUN if [ "$TARGETARCH" = "amd64" ] ; then \
+    curl -sL -o /tmp/mold.tar.gz "https://github.com/rui314/mold/releases/download/v${MOLD_VERSION}/mold-${MOLD_VERSION}-x86_64-linux.tar.gz" \
+    && echo "${MOLD_SHA256_AMD64}  /tmp/mold.tar.gz" | sha256sum --check \
+    && tar -C /usr/local --strip-components=1 -xzf /tmp/mold.tar.gz \
+    && rm /tmp/mold.tar.gz ; fi
+
```
Ishirui, line 67: IMO the cleaner way to handle this is to declare an intermediate stage that only gets run if the target arch is appropriate, something like this:
```dockerfile
FROM <previous stage> AS build_mold_arm64
ADD ..._aarch64 --checksum ...

FROM <previous stage> AS build_mold_amd64
ADD ..._x86_64 --checksum ...

FROM <previous stage> AS final
COPY --from build_mold_${TARGETARCH} ...
```
But this is also quite clunky haha, and I see you've copied the pattern just above so no big deal

Ishirui (approved):
Clean !

devcontainer/Dockerfile
pgimalac: Yeah I tried to stay close to existing patterns 😄

aiuto (approved):

PR #1142, opened 2026-04-13
Add explicit memory requests and limits for buildx

pgimalac:
### What does this PR do?
Add explicit memory requests and limits for buildx driver / service in build jobs.

### Motivation
Fix jobs hanging because the remote builder gets OOM-killed

### Additional Notes
Triggered some pipelines while disabling cache, didn't get any OOM / failure, so at least it's not worst.


pgimalac:
This doesn't work, couldn't get docker in docker to work, and couldn't set arbitrary memory requests on the buildx builder...

PR #1145, opened 2026-04-20
Create a flavor of the dev environment that works with workspaces

KevinFairise2:
### What does this PR do?

Create a new flavor of our devenv. That will be compatible with Datadog workspaces.
Creating a multi layer image, the devenv will configure ssh to work with our dda command. The other one will create required user so that it works in workspaces

### Motivation

Allow developers to use the dev environment as a workspaces devcontainer


dev-envs/linux/dda.sh
```diff
@@ -2,12 +2,15 @@
...
-else
-    ln -s "$(/root/miniforge3/condabin/conda run -n ddpy3 which dda)" /usr/local/bin/dda
-fi
+ARCH=$(dpkg --print-architecture) && \
+    if [ "$ARCH" = "amd64" ]; then \
+    GO_ARCH="x86_64"; \
+    elif [ "$ARCH" = "arm64" ]; then \
+    GO_ARCH="aarch64"; \
+    fi && \
+    curl -fsSL --retry 4 "https://github.com/DataDog/datadog-agent-dev/releases/download/${DDA_VERSION}/dda-${GO_ARCH}-unknown-linux-gnu.tar.gz" \
```
aiuto: Drive-by observation.
Why do we publish some things based on "GO_ARCH" instead of the arch names we use for our other products. 
No one cares if a tool is written in Go or Rust or Fortran. They only care about the architecture of the machine they are using it from.

DO NOT LET THAT THOUGHT DERAIL THIS PR. It's just a thought.

KevinFairise2: Unfortunately `arm64` and `aarch64` is used a bit interchangeably everywhere, so we need to conversion layer here.
I agree that `GO_ARCH` was maybe not the best naming for that variable and `DDA_ARCH` would be better.

.gitlab/build.yml
```diff
@@ -195,6 +196,29 @@ build_dev_env_linux_arm64:
     IMAGE: dev-env-linux-$DD_TARGET_ARCH
     BASE_IMAGE: registry.ddbuild.io/ci/datadog-agent-buildimages/linux$ECR_TEST_ONLY
     BASE_IMAGE_TAG: $IMAGE_VERSION-arm64
+    TARGET: devenv
+
+build_dev_env_linux_workspace_x64:
```
Ishirui: nit: maybe calling it a "workspace env" is a bit shorter, so `build_workspace_env_x64` (linux is the only platform anyway right ?)

.gitlab/build.yml
```diff
@@ -195,6 +196,29 @@ build_dev_env_linux_arm64:
...
+build_dev_env_linux_workspace_arm64:
+  extends: [.fast_build, .no_schedule, .arm]
+  needs:
+    - job: build_linux_arm64
+  variables:
+    DOCKERFILE: dev-envs/linux/Dockerfile
+    IMAGE: dev-env-linux-workspace-$DD_TARGET_ARCH
+    BASE_IMAGE: registry.ddbuild.io/ci/datadog-agent-buildimages/linux$ECR_TEST_ONLY
+    BASE_IMAGE_TAG: $IMAGE_VERSION-arm64
+    TARGET: workspace
```
Ishirui, line 221: nit: maybe we could use a template job ?

dev-envs/linux/zsh.sh
```diff
@@ -27,7 +27,7 @@ fi
 
 tar -xf "${workdir}/${archive_name}" -C "${workdir}" --strip-components 1
 pushd "${workdir}"
-./configure --with-tcsetpgrp
+./configure --with-tcsetpgrp --sysconfdir=/etc/zsh --enable-etcdir=/etc/zsh
```
KevinFairise2, line 30: I think that one is harmless. And is required for zsh to work properly in workspaces. They expect zsh to source zshenv in /etc/zsh/zshenv. Not having that break a lot of things that are provided for us by workspaces

.gitlab/release.yml
```diff
@@ -1,4 +1,65 @@
...
+    - !reference [.build, before_script]
+  script:
+    - >
+      docker manifest create $TARGET_IMAGE:$IMAGE_VERSION
+      $AMD64_IMAGE:$IMAGE_VERSION
+      $ARM64_IMAGE:$IMAGE_VERSION
+    - MANIFEST=$(docker manifest inspect $TARGET_IMAGE:$IMAGE_VERSION)
+    - DIGEST=sha256:$(echo -n "$MANIFEST" | sha256sum | awk '{print $1}')
+    - docker manifest push $TARGET_IMAGE:$IMAGE_VERSION
+    - ddsign sign $TARGET_IMAGE:$IMAGE_VERSION@${DIGEST}
```
Ishirui: We should avoid duplicating code - this is very similar to the already existing `build_multiarch` job. We could probably extract some of this to a reused template.

So:
- `build_multiarch` becomes `.build_multiarch`
- The main `linux` multiarch images build job gets renamed to `build_linux_multiarch` and extends `.build_multiarch`
- We create jobs for the multiarch dev env and workspace env build jobs (`build_dev_env` and `build_workspace_env`) that extend `.build_multiarch` specifying some variables (`DOCKERFILE`...) and parent jobs

dev-envs/linux/Dockerfile
```diff
@@ -40,18 +40,38 @@ RUN --mount=type=bind,src=./starship.sh,dst=/mnt/starship.sh /mnt/starship.sh
...
 # Run steps that must happen last
 RUN --mount=type=bind,src=./finalize.sh,dst=/mnt/finalize.sh /mnt/finalize.sh
 ENV LC_ALL=C.UTF-8
 ENV LANG=C.UTF-8
+# Create the dog group and user
+RUN set -x \
+    && groupadd --gid 501 dog  \
+    && useradd --gid 501 --uid 501 --create-home --groups users dog
+# MCP
+EXPOSE 9000
```
Ishirui: We can put this in the `workspace` specific stage

dev-envs/linux/Dockerfile
```diff
@@ -40,18 +40,38 @@ RUN --mount=type=bind,src=./starship.sh,dst=/mnt/starship.sh /mnt/starship.sh
...
 COPY entrypoint.sh /entrypoint.sh
 ENTRYPOINT ["/entrypoint.sh"]
+
+# ---- workspace: no SSH, no AWS CLI ----
+FROM base AS workspace
+
+# Remove AWS CLI v2 (not needed in dev-env)
+# https://docs.aws.amazon.com/cli/latest/userguide/uninstall.html
+RUN rm -f /usr/local/bin/aws /usr/local/bin/aws_completer \
+    && rm -rf /usr/local/aws-cli
```
Ishirui: Does it actually matter if it's there ? (+ comment is wrong, should probably be `not needed in workspace env`)

dev-envs/linux/dda.sh
```diff
@@ -2,12 +2,15 @@
...
+    | tar -xzf - -C /usr/local/bin && \
+    chmod +x /usr/local/bin/dda
```
Ishirui, line 13: I think this means that `dda` will install itself on first run. Maybe it would be better to just have a fully-installed version baked in the image ? Why not keep the pip-based installation in that case ?

I know this sometimes broke with workspaces, so if you've tried that and it doesn't work, disregard 😅 

Also keep in mind that `dda` is probably preinstalled in the base `linux` CI image, so a blind reinstall is probably not a great idea. `pip install --upgrade` would be better imo

dev-envs/linux/Dockerfile
KevinFairise2: Yes it matters, it actually makes the base feature installation crash. I will look into making their feature more permissive and not hard fail when it is already there. But to have it working now that's easier for us to remove it on our end

dev-envs/linux/zsh.sh
aiuto: FWIW... in any environment (native or container), I have to import my personal set of fixes to .rc files to get zsh to behave the way I want it to. So, what is provided by default is almost immaterial to me - as a dev, I generally wipe out /etc/zsh and override with my personal shell behavior.

KevinFairise2: I think they support adding you own dotfiles. But you're probably gonna have a workspace working as you expect if you delete the /etc/zsh folder entirely, they rely on that to source some environment variable when you create the SSH session

dev-envs/linux/dda.sh
aiuto: This can open a bigger discussion.  My observation is that we have toil and churn from not versioning developer images with the release branches.  dda is part of this because it has behaviors that need to change with the code, and behaviors that neet to change with the infrastructure. I think those need to be distinct tools.
Perhaps we can do a meeting about it this week. I'll get up extra early for it.

Again.  This is not a comment you need to resolve. Just a point.

.gitlab/build.yml
KevinFairise2: I think we can, but the jobs is already pretty factorized, we only pass variables to configure it. I could add one more layer of factorization but I feel like it would make it less readable

.gitlab/release.yml
KevinFairise2: Removed those testing pieces.

dev-envs/linux/dda.sh
KevinFairise2: I am not sure I will be able to resolve that issue in the PR.
I though about several things. The requirement I have is having dda installed not in a python environment owned by `root`. Otherwise I will not be able to use it when running in workspaces as the `bits` user.
I could modify it to `pip install` it as `bits` user but I think the python environment is managed currently with `conda` for the root user so that would probably not works well without refactoring completely the way we manage python in the base build image.
The options I have for now:
- Keep it as is. I tested it it works fine in workspaces and does not break the dev-env
- Remove everything I did on dda in that PR and instead move the installation of dda in the workspaces in a `devcontainer` feature inside `datadog-agent`

I think I prefer keeping it in the PR. And I think in any case we will end up with two `dda` being installed in two different places for the different users

Ishirui: Could you create a ticket to track this ? It's not super urgent to fix but it would be nice to not forget

Ishirui (approved):
🚀

dev-envs/linux/dda.sh
aiuto: One of them on the build side is https://datadoghq.atlassian.net/browse/ABLD-359.
But there really should be an overall tracking ticket in https://datadoghq.atlassian.net/jira/core/projects/SPPL 
to capture the concept of decoupling the things that vary by code branch/release from the things that vary because our infrastructure changes. 
When they are in the same file (source or binary), reverting "just a piece" ™ is more difficult than it should be.

PR #1150, opened 2026-04-24
migrate docker_x64 build to bake

chouquette:
### What does this PR do?

Migrate `docker_x64` build to bake

### Motivation

While working on https://github.com/DataDog/datadog-agent-buildimages/pull/1149 I realized that there were some discrepancies between the `docker_x64` image build and the most of the build images, which made the change a bit harder that I was expecting.

Migrating to bake means we will share the same tooling and updates to the bake file, without needing to add more targets to renovate


chouquette:
@codex review

.gitlab/build.yml
```diff
@@ -95,10 +95,12 @@ build_dd_agent_testing:
     IMAGE: dd-agent-testing
 
 build_docker_x64:
-  extends: [.fast_build, .always, .x64]
+  extends: [.build_linux, .always, .x64]
```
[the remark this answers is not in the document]
Ishirui: This is a valid concern - sometime in Q2 unsigned images will be prevented from use in CI configs. The `linux` images currently do this when they are merged into a single multiarch image in `.build_multiarch` - I think you have two options here: either add `ddsign sign $TARGET_IMAGE:$IMAGE_VERSION@${DIGEST}` to the build jobs for this image, or make a multiarch variant that also goes through the `.build_multiarch` job (like the main `linux` images).

I'd prefer the second option, but that requires a companion PR to `datadog-agent` so the first one is easier in the meantime

docker-bake.hcl
```diff
@@ -324,3 +329,46 @@ target "linux-arm64-ci_test_only" {
...
+  context    = "./"
+  cache-from = [docker_x64_cache_details_main]
+  tags       = ["${repo_name}/docker_x64:latest"]
+  args       = merge(versions, go_versions, go_checksums_amd64, { BUILDENV_REGISTRY = BUILDENV_REGISTRY })
+}
+
+target "docker_x64-amd64" {
+  inherits  = ["_fake_docker_x64-local"]
+  platforms = ["linux/amd64"]
+}
```
Ishirui: This feels unnecessary, and more importantly quite confusing - `docker_x64` already include the architecture in the image name itself, and only the `amd64` flavor exists anyway.
My preferred fix would be to merge `docker_x64` and `docker_arm64` into a single Dockerfile and do a multiarch build (like for `linux`), but this is a much larger piece of work.

It's especially weird because the `docker_arm64` image lacks a lot of stuff now in the `x64` one... feels poorly maintained :/

docker-x64/Dockerfile
```diff
@@ -1,18 +1,23 @@
+ARG DD_OCTO_STS_VERSION=latest
 ARG BUILDENV_REGISTRY
+
+FROM registry.ddbuild.io/dd-octo-sts:${DD_OCTO_STS_VERSION} AS dd_octo_sts_builder
+# Dummy stage to download dd-octo-sts as it is distributed as a container image
+
 FROM ${BUILDENV_REGISTRY}/images/docker:27.3.1
 
 ARG PYTHON_VERSION=3.12.6
```
Ishirui: This should be `PY3_VERSION` so it can be handled via the bake file as well

docker-x64/Dockerfile
```diff
@@ -1,18 +1,23 @@
...
 ARG DDA_VERSION
-ARG DDA_NO_DYNAMIC_DEPS
+ARG DDA_NO_DYNAMIC_DEPS=1
 ARG GO_VERSION
-ARG GO_SHA256_LINUX_AMD64
+ARG GO_SHA256
 ARG DATADOG_CI_VERSION=5.12.1
```
Ishirui: nit: better to avoid setting default versions for stuff specified in the bakefile, to match the main `linux` image

.gitlab/build.yml
Ishirui: nit: would be nice to rename `.build_linux` to `.build_bake` or something, since it's not tied exclusively to the `linux` image anymore

Ishirui (changes requested):
Thanks a lot for taking the time to work on this ! Finally I'm not the only one using bake 😂 

I'm requesting changes only for the signing piece, we don't want `datadog-agent` CI suddenly starting to fail once they start enforcing image signing.

Otherwise the changes look good ! A bigger endeavour would be to merge `docker_x64` and `docker_arm64` into a single Dockerfile, and do a multiarch build - similar to what is done for `linux`. That would be a lot cleaner and simpler to manage. There is no reason those two images should be built from different Dockerfiles - it's especially bad now that the `x64` has some extra stages that don't exist in the `arm64`. But this can be done at a later date.

docker-bake.hcl
```diff
@@ -324,3 +329,46 @@ target "linux-arm64-ci_test_only" {
...
+  default = "type=registry,ref=${registry_name}/${repo_name}/docker_x64:${cache_key_branch}"
+}
+
+target "_fake_docker_x64-local" {
+  dockerfile = "docker-x64/Dockerfile"
+  context    = "./"
+  platforms  = ["linux/amd64"]
+  cache-from = [docker_x64_cache_details_main]
+  tags       = ["${repo_name}/docker_x64:latest"]
+  args       = merge(versions, go_versions, go_checksums_amd64, { BUILDENV_REGISTRY = BUILDENV_REGISTRY })
```
Ishirui: You need `checksum_amd64` here as well, otherwise none of the arch-specific checksums will be passed

Ishirui (approved):
See comment for getting the build to pass :)
Otherwise LGTM !

PR #1152, opened 2026-04-28
fix(docker_x64): Revert tagging scheme to original `v$CI_PIPELINE_ID-${CI_COMMIT_SHORT_SHA}`

Ishirui:
### What does this PR do?
Revert tagging scheme to original `v$CI_PIPELINE_ID-${CI_COMMIT_SHORT_SHA}`

### Motivation
After #1150 the tagging scheme was changed to `v$CI_PIPELINE_ID-${CI_COMMIT_SHORT_SHA}-amd64`, which would require a companion change on `datadog-agent`. Moreoever it makes no sense since the arch is already included in the image name itself.

### DIFF
--- a/.gitlab/build.yml
+++ b/.gitlab/build.yml
@@ -102,7 +102,7 @@ build_docker_x64:
       aud: image-integrity
   variables:
     IMAGE: docker_x64
-    IMAGE_VERSION: $IMAGE_VERSION_ROOT-amd64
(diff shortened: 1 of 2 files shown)


rdesgroppes (approved):
for the record, see Slack [discussion thread](https://dd.slack.com/archives/C06PBHLD4DQ/p1777379268928249) - well, until it vanishes...

KevinFairise2 (approved):

chouetz (approved):

PR #1173, opened 2026-05-20
rpm-armhf: install libxslt-devel

chouquette:
### What does this PR do?

Install libxslt-devel

### Motivation

This is needed to bootsrap dda on the build image


aiuto (approved):
ISTM that we need a project to cross compile all the deps for armhf.

PR #1174, opened 2026-05-21
Handle existing host GID in dev env entrypoint

quentinus95:
### What does this PR do?

The dev environment entrypoint maps the container user to the host UID/GID so mounted files keep the expected ownership. On macOS the primary group is commonly staff with GID 20, which already exists in the base image as dialout. Creating a new group with that GID fails and prevents sshd from starting.

Today, we only check if `TARGET_USER` exists before creating *the group* and the user.

Check whether the numeric host GID already exists before calling groupadd, then pass the numeric GID to useradd. This preserves the host GID mapping while avoiding the startup failure.

### Motivation

Current test running:

```sh
dda --version
dda, version 0.34.0
dda env dev remove
dda env dev start
Waiting for container: dda-linux-container-default
[...]
RuntimeError: Container `dda-linux-container-default` is not ready
[...]
+[08:04:58.575] groupadd -g 20 dd
groupadd: GID '20' already exists
```

### DIFF
--- a/dev-envs/linux/init/entrypoint.sh
+++ b/dev-envs/linux/init/entrypoint.sh
@@ -23,8 +23,10 @@ if [[ ! -f "${startup_indicator}" ]]; then
         TARGET_UID="${HOST_UID:-1002}"
         TARGET_GID="${HOST_GID:-${TARGET_UID}}"
 
-        # Create primary user and group
-        groupadd -g "${TARGET_GID}" "${TARGET_GROUP}"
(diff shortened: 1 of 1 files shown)


quentinus95:
Ran `dda env dev start --image registry.ddbuild.io/ci/datadog-agent-buildimages/dev-env-linux-aarch64_test_only:v114424095-be766a1a`, alles gut

KevinFairise2 (approved):

PR #1177, opened 2026-05-22
Improve python environment management for workspaces, add back claude in dev env

KevinFairise2:
### What does this PR do?

Improve python environment handling, when logged in as bits user in workspaces it should also enable python conda environment. So that everything works as expected, like in the CI
Also add a dotslash install of Claude

### DIFF
--- a/dev-envs/linux/variants/workspaces/setup.sh
+++ b/dev-envs/linux/variants/workspaces/setup.sh
@@ -41,7 +41,8 @@ cat >/etc/sudoers.d/95-dog-nopasswd <<'EOF'
 EOF
 chmod 0440 /etc/sudoers.d/95-dog-nopasswd
 
-su bits -c "conda init --all"
+
+su bits -c "${DD_BUILD_INSTALL_ROOT}/conda/condabin/conda init --all"
 
 echo "conda activate ddpy3" >> /home/bits/.zshrc
 # zsh does not source /etc/profile.d by default. Workspaces and the Agent
--- a/dev-envs/linux/variants/workspaces/setup.sh
+++ b/dev-envs/linux/variants/workspaces/setup.sh
@@ -41,6 +41,9 @@ cat >/etc/sudoers.d/95-dog-nopasswd <<'EOF'
 EOF
 chmod 0440 /etc/sudoers.d/95-dog-nopasswd
 
+su bits -c "conda init --all"
+
+echo "conda activate ddpy3" >> /home/bits/.zshrc
 # zsh does not source /etc/profile.d by default. Workspaces and the Agent
 # feature both write *-workspace-env.sh snippets there.
 install -d /etc/zsh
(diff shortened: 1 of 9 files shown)


dev-envs/linux/variants/workspaces/setup.sh
```diff
@@ -41,6 +41,10 @@ cat >/etc/sudoers.d/95-dog-nopasswd <<'EOF'
 EOF
 chmod 0440 /etc/sudoers.d/95-dog-nopasswd
 
+
+su bits -c "${DD_BUILD_INSTALL_ROOT}/conda/condabin/conda init --all"
+
+echo "conda activate ddpy3" >> /home/bits/.zshrc
```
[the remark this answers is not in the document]
KevinFairise2: In workspaces we expect user to login in zsh only, we can add more later on if needed

tools/dotslash/config/aws-vault.json
aiuto, line 1: Just out of curiousity, will these land in the developer and CI images?
We can read them from bazel repository rules to make these tools available to bazel actions.
We can pull the digest and URL right from these files so there is nothing to keep in sync in MODULE.bazel - that's a big maintainability win there. Bazel will end up downloading them (as needed on demand) to repository cache, but that's fine for most things.  As we complete the move to bazel build, there will be very few pure shell invocations that would invoke dotslash from CI.

aiuto (approved):

tools/dotslash/config/aws-vault.json
KevinFairise2: Dotslash is used only in developer environment image. The way it works is that it does not ship anything inside the image. The binaries are automatically downloaded the first time they are called

aiuto: > Dotslash is used only in developer environment image.

Excellent.

> The way it works is that it does not ship anything inside the image. The binaries are automatically downloaded the first time they are called

That's what I expect.  But here's the hypothetical that Ofek and I discussed earlier this week.
Bazel projects often need a way to grab a prebuilt too in a safe way.  We do it with a package called multitool.  https://github.com/DataDog/datadog-agent/blob/e21494b2398fee15b34296f096baaab589105b85/MODULE.bazel#L157 and
https://github.com/DataDog/datadog-agent/blob/main/bazel/prebuilt_buildtools.json

The data in the json files is essentially the same as the .dotslash files. So we could easily make something like multitools that drives from the dotslash files so a single definition is usable by all three of 
- developers using dotslash
- bazel rules that need the tool (in CI or developer builds, that doesn't matter)
- `bazel run` commands where you want to script using the tool, but don't have a rule (buildozer is a common tool in that situation)

What I'm getting at is that we can share the declarations to prevent us getting out of alignment.
Of course, we would want to move some of the dotslash definitions into datadog-agent, so we could rev them without having to update build image.

PR #1178, opened 2026-05-25
Add more tools

ofek:
### What does this PR do?

This is a standalone PR including tools from https://github.com/DataDog/datadog-agent-buildimages/pull/1177 and others that are useful.

### Additional Notes

- Fixed tooling to support DotSlash config with non-read-only artifacts
- Updated Nushell and Starship installation logic
- Minor updates to various tooling configuration defaults


aiuto (approved):

dev-envs/linux/zsh.sh
```diff
@@ -39,13 +39,13 @@ OH_MY_ZSH_COMMIT="61bacd95b285a9792a05d1c818d9cee15ebe53c6"
...
 
-# Fix locale broken by Oh My Zsh:
-# https://github.com/starship/starship/issues/2176#issuecomment-1783086362
-umask 0002
-cat <<'EOF' >> "${HOME}/.zshrc"
-export LC_ALL="C.UTF-8"
-export LANG="C.UTF-8"
+    # Set up Starship for zsh.
+    cat <<'EOF' >> "${HOME}/.zshrc"
+eval "$(starship init zsh)"
```
aiuto: This should be opt in... messing around with the tools people expect from the shell is a touchy subject.

dev-envs/linux/zsh.sh
```diff
@@ -39,13 +39,13 @@ OH_MY_ZSH_COMMIT="61bacd95b285a9792a05d1c818d9cee15ebe53c6"
...
     umask 0002
     sh -c "$(curl -fsSL https://raw.githubusercontent.com/ohmyzsh/ohmyzsh/${OH_MY_ZSH_COMMIT}/tools/install.sh)"
-)
 
-# Fix locale broken by Oh My Zsh:
-# https://github.com/starship/starship/issues/2176#issuecomment-1783086362
-umask 0002
-cat <<'EOF' >> "${HOME}/.zshrc"
-export LC_ALL="C.UTF-8"
-export LANG="C.UTF-8"
```
aiuto, line 49: LOL.. You are removing the things I always set.
OTOH.... I use custom RCs for everything, so this is sort of moot.

ofek: This wasn't removed but rather placed at the very start of the Dockerfile to have consistency among shells.

image

dev-envs/linux/zsh.sh
ofek: Great catch, thanks! I forgot to push a change I had intended.

aiuto (approved):
Still LGTM

chouetz (approved):

PR #1183, opened 2026-06-01
Correct permissions of the home directory in the Linux dev env

ofek:
### What does this PR do?

- Ensure that the home directory of the runtime-created user has the expected permissions
- Fix warnings being emitted from Zsh

### Motivation

- There's no reason to have the user's access to its home directory granted via group membership rather than normal ownership since the root user wouldn't write to the directory at runtime. The only thing to look out for is that the home directory within the image stays barebones so that the `chown` at first startup is fast.
- When using Zsh there was a warning during initialization from Oh My Zsh and another upon exit from Zsh itself.

```
[oh-my-zsh] Insecure completion-dependent directories detected:
drwxrwsr-x 12 root build-shared 4096 May 28 15:57 /home/dd/.oh-my-zsh
drwxrwxr-x  4 root root         4096 May 27 14:37 /home/dd/.oh-my-zsh/custom

[oh-my-zsh] For safety, we will not load completions from these directories until
[oh-my-zsh] you fix their permissions and ownership and restart zsh.
[oh-my-zsh] See the above list for directories with group or other writability.

[oh-my-zsh] To fix your permissions you can do so by disabling
[oh-my-zsh] the write permission of "group" and "others" and making sure that the
[oh-my-zsh] owner of these directories is either root or your current user.
[oh-my-zsh] The following command may help:
[oh-my-zsh]     compaudit | xargs chmod g-w,o-w

[oh-my-zsh] If the above didn't help or you want to skip the verification of
[oh-my-zsh] insecure directories you can set the variable ZSH_DISABLE_COMPFIX to
[oh-my-zsh] "true" before oh-my-zsh is sourced in your zshrc file.
```

```
zsh: rewriting /home/dd/.zsh_history would change its ownership -- skipped
```

### Additional Notes

This also installs Oh My Zsh on-demand rather than being shipped with the image and persists the installation for the user so that their changes aren't tied to container lifecycles

### DIFF
--- a/dev-envs/linux/Dockerfile
+++ b/dev-envs/linux/Dockerfile
@@ -14,8 +14,8 @@ USER root
 ENV LC_ALL=C.UTF-8
 ENV LANG=C.UTF-8
 
-# Set up HOME as root-owned and build-shared writable so host-mapped users
-# with unknown UID/GID can use it without chown.
+# Set up HOME during the image build. The entrypoint assigns it to the
+# runtime target user before startup creates user state.
 ENV HOME=/home/dd
 ENV PATH="${HOME}/.local/bin:${PATH}"
 RUN mkdir -p "${HOME}" && \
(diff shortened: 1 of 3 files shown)


quentinus95 (approved):
Tested with `dev-env-linux-x64_test_only:v116146686-6026929a` 👍

PR #1190, opened 2026-06-04
[ACIX-1444] Install and configure Claude and Codex in developer environment

KevinFairise2:
### What does this PR do?

Install and configure Claude and Codex for use in developer environment

### Motivation

Make it easier to work with Coding agent inside developer environment

Note: For now all the MCP that rely on browser redirect authentication are hard to configure, because we do not have a mechanism to forward those browser request to the local laptop. Everything will work out of the box in workspaces because there is a mechanism for that

### DIFF
--- a/dev-envs/linux/claude.sh
+++ b/dev-envs/linux/claude.sh
@@ -8,6 +8,25 @@ umask 0002
 export CLAUDE_CONFIG_DIR="${DD_BUILD_CONFIG_ROOT}/claude"
 mkdir -p "${CLAUDE_CONFIG_DIR}"
 
+cat <<'EOF' > "${CLAUDE_CONFIG_DIR}/settings.json"
+{
+  "apiKeyHelper": "ddtool auth token rapid-ai-platform --datacenter us1.ddbuild.io",
+  "extraKnownMarketplaces": {
+    "datadog": {
+      "source": {
+        "source": "github",
+        "repo": "DataDog/claude-marketplace"
+      }
+    }
+  },
+  "env": {
+    "ANTHROPIC_BASE_URL": "https://ai-gateway.us1.ddbuild.io",
+    "ANTHROPIC_CUSTOM_HEADERS": "source: claude-code\norg-id: 2\nprovider: anthropic\nclaude-code: true",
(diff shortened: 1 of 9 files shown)


tools/dotslash/config/codex.json
```diff
@@ -0,0 +1,41 @@
...
+        "uncompressed_size": 234632293
+      }
+    }
+  },
+  "name": "codex",
+  "platforms": {
+    "linux-aarch64": {
+      "digest": "fb2eeeb03c48950fb361c77d8a65c32b56aafdc1a1cc68e3ebd7b0db2c5ef117",
+      "format": "tar.zst",
+      "hash": "blake3",
```
aiuto, line 17: WTF is blake3?  Why not sha256, which is on all the machines?

KevinFairise2: Hmm I think this is the prefered algorithm for dotslash. And codex actually give the manifest to use in their releases: https://github.com/openai/codex/releases/download/rust-v0.138.0-alpha.6/codex

aiuto: It's just annoying that we don't have that on the mac machines by default.

aiuto (approved):

dev-envs/linux/claude.sh
```diff
@@ -2,4 +2,17 @@
...
+export CLAUDE_CONFIG_DIR="${DD_BUILD_CONFIG_ROOT}/claude"
+mkdir -p "${CLAUDE_CONFIG_DIR}"
+
+# `claude` comes from tools.sh (DotSlash). Add MCP servers for the image here, for example:
+claude mcp add -s user --transport http "ddci-mcp-prod" \
+  'https://ddci-mcp.mcp.us1.ddbuild.io/internal/mcp'
+
+claude mcp add datadog-google-workspace -s user --transport http https://google-workspace-mcp-server-834963730936.us-central1.run.app/mcp
+
+claude mcp add --transport http -s user atlassian https://mcp.atlassian.com/v1/mcp
```
chouetz, line 40: Silly question, I suppose `datadog-mcp` is also installed, why don't we see it here? And should we plug also slack?

chouetz (approved):

PR #1194, opened 2026-06-09
Restore compatibility between our workspace flavor and workspace base feature

KevinFairise2:
### What does this PR do?

Remove a lot of things that were added with the goal of completely managing all the provisioning without relying on the base feature. I think this is not a good idea and will break as soon as the provisioning changes.
Instead we decide to move with an approach where:
- We install all our tooling in the base image, in non-standard places 
- We let the base feature do all the provisioning, potentially installing tools that we already manage, removing some tools that hard conflicts with ours
- We then update the paths in the image to make sure that our tooling is used first, and that the `bits` user is authorized to write/read on our paths by adding it to the `build-shared` group

The main advantage is that we can target our tooling, but we also benefits from things that workspaces are doing well in their environment. Setting up commit signing properly, login browser forwarding, keeping a way for us to benefits from future improvement. Any tooling that we did not install and that workspaces decide to provide will be made available automatically as well

Goes with: https://github.com/DataDog/datadog-agent/pull/51995

### Motivation

Find a path forward to use our developer environment in workspaces, without taking too much risk on provisioning breaking us


dev-envs/linux/Dockerfile
```diff
@@ -16,7 +16,7 @@ ENV LANG=C.UTF-8
 
 # Set up HOME during the image build. The entrypoint assigns it to the
 # runtime target user before startup creates user state.
-ENV HOME=/home/dd
+ARG HOME=/home/dd
```
KevinFairise2, line 19: We do not want to persist it in the image. Otherwise when the provisioning is done with workspaces base feature, we get some scripts executed as rool with that HOME and it breaks

dev-envs/linux/Dockerfile
```diff
@@ -83,21 +83,15 @@ ENV DD_TARGET_GROUP=dog
...
 RUN --mount=type=bind,src=./variants/workspaces/setup.sh,dst=/mnt/workspaces-setup.sh \
     /mnt/workspaces-setup.sh
-COPY --chmod=755 variants/workspaces/adduser.sh /opt/doghome/sbin/adduser.sh
 
-# Run workspace sessions as the pre-created Workspaces SSH user.
-ENV HOME=/home/bits
-ENV IN_WORKSPACE=1
-ENV XDG_BIN_HOME="${HOME}/.local/bin"
-ENV PATH="${XDG_BIN_HOME}:${PATH}"
+RUN rm -f /usr/local/bin/aws /usr/local/bin/aws_completer \
```
KevinFairise2, line 87: Still conflicting with base feature, let's remove it and let workspace install it

dev-envs/linux/Dockerfile
```diff
@@ -83,21 +83,15 @@ ENV DD_TARGET_GROUP=dog
...
-# Set up OpenSSH server
-# IMPORTANT: make sure this comes last so Git operations don't use SSH during building
-RUN --mount=type=bind,src=./ssh.sh,dst=/mnt/ssh.sh /mnt/ssh.sh
-RUN printf '%s\n' \
-    'StreamLocalBindUnlink yes' \
-    'PasswordAuthentication no' \
-    >> /usr/local/etc/sshd_config
+
+# Remove ddtool from the image to avoid conflicts with Workspace one.
+RUN rm -f /usr/local/bin/ddtool
```
KevinFairise2, line 92: Removing that one as it is installed by workspace and needed during the provisioning, Keeping it breaks provisioning

dev-envs/linux/Dockerfile
```diff
@@ -83,21 +83,15 @@ ENV DD_TARGET_GROUP=dog
...
+# Unset XDG_DATA_HOME to avoid conflicts during the workspace provisioning.
+ENV XDG_DATA_HOME=
```
KevinFairise2, line 94: It will be set again automatically when the user logins thanks to: https://github.com/DataDog/datadog-agent-buildimages/pull/1194/files#diff-e1f8f6fc8265462b83c5e9d5e553c8e2e2b567a02cd506c75b4bc4e4484e100dR11-R13

dev-envs/linux/Dockerfile
```diff
@@ -83,21 +83,15 @@ ENV DD_TARGET_GROUP=dog
...
+RUN rm -f /usr/local/bin/aws /usr/local/bin/aws_completer \
+    && rm -rf /usr/local/aws-cli
 
-# Set up OpenSSH server
-# IMPORTANT: make sure this comes last so Git operations don't use SSH during building
-RUN --mount=type=bind,src=./ssh.sh,dst=/mnt/ssh.sh /mnt/ssh.sh
-RUN printf '%s\n' \
-    'StreamLocalBindUnlink yes' \
-    'PasswordAuthentication no' \
-    >> /usr/local/etc/sshd_config
```
KevinFairise2, line 100: Workspace will handle ssh

dev-envs/linux/variants/workspaces/setup.sh
```diff
@@ -7,49 +7,31 @@ bits_home="/home/bits"
...
 groupadd --gid 501 dog
 useradd --gid dog --uid 501 --home-dir "${dog_home}" --shell /bin/bash --groups users,build-shared,sudo dog
-useradd --gid dog --uid 2000 --home-dir "${bits_home}" --shell /usr/local/bin/zsh --groups users,build-shared,sudo bits
-if getent group docker >/dev/null; then
-    usermod -a -G docker bits
-fi
+    
+while IFS= read -r line; do
+    printf 'export %s=%q\n' "${line%%=*}" "${line#*=}"
+done < <(env | grep -Ev "^(HOME=|USER=|MAIL=|LS_COLORS=|HOSTNAME=|PWD=|TERM=|SHLVL=|LANGUAGE=|_=)") > /etc/profile.d/dd-agent-workspace-env.sh
```
KevinFairise2, line 14: We no longer create the bits user. The base feature do it and we add it in the build-shared group later on

dev-envs/linux/variants/workspaces/setup.sh
```diff
@@ -7,49 +7,31 @@ bits_home="/home/bits"
...
-install -d /etc/zsh
-cat >/etc/zsh/zshenv <<'EOF'
-if [[ -z "$PATH" || "$PATH" == "/bin:/usr/bin" ]]; then
-    export PATH="/usr/local/bin:/usr/bin:/bin:/usr/games"
-fi
-
-for s in $(find /etc/profile.d/ -type f -name "*-workspace-env.sh" | sort); do
-    source "$s"
-done
-EOF
```
KevinFairise2, line 55: Already done by the base feature

chouetz: Should we keep some comments in the code about these?

dev-envs/linux/Dockerfile
```diff
@@ -106,7 +100,7 @@ RUN --mount=type=bind,src=./finalize.sh,dst=/mnt/finalize.sh /mnt/finalize.sh
 EXPOSE 22
 
 # Run SSH server
-CMD ["/usr/local/sbin/sshd", "-D", "-e"]
+CMD ["sleep", "infinity"]
```
chouetz, line 103: 🛌🏽 😴

dev-envs/linux/variants/workspaces/setup.sh
chouetz: Then I suppose you need to remove `build-shared` as requested [here](https://github.com/DataDog/datadog-agent-buildimages/pull/1194/changes#r3505997332)?

chouetz (approved):
I suppose you need to check codex comment, otherwise LGTM

dev-envs/linux/variants/workspaces/setup.sh
```diff
@@ -7,49 +7,31 @@ bits_home="/home/bits"
 
 groupadd --gid 501 dog
 useradd --gid dog --uid 501 --home-dir "${dog_home}" --shell /bin/bash --groups users,build-shared,sudo dog
```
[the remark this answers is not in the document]
KevinFairise2: It is done in the feature on datadog-agent side

dev-envs/linux/variants/workspaces/setup.sh
KevinFairise2: For that one specifically, I do not think we should comment code that is not there. And it has been added recently but was never used in fact

dev-envs/linux/variants/workspaces/setup.sh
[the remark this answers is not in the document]
KevinFairise2: Actually startup.sh is no longer called, in workspaces. The entrypoint is overriden

dev-envs/linux/variants/workspaces/setup.sh
KevinFairise2: Hmm I tried to remove the `dog` user as well but we cannot, the base feature expect it to be there. And I think the Codex comment is not really accurate. It says that the bits user will no longer have the build-shared group which is true but this is something we handle in the steps that are executed after the workspace is created, on datadog-agent side

PR #1207, opened 2026-06-15
[AGENTRUN-1173] fix(ci): target other branches than main in agent repo when backporting go update

Enzu83:
### What does this PR do?

Fixes a use case in the Go update PR automation process when backporting Go update by checking the name of the targeted branch in the commit name alongside the trigger `[push-to-datadog-agent]`.

Commits with names like "[push-to-datadog-agent][7.80.x] ..." will trigger the PR automation process.

The Github action to create a Go update PR has been updated to accept the targeted branch as an optional input.

### Motivation

The job `push-to-datadog-agent` creates a PR in datadog-agent automatically when the Go update Github action is run. However, it targets the wrong branch when backporting a Go update.

Testing isn't trivial since the feature is based on commit names, CI, automatically opened PRs.
The job retrieves the correct branch name at least.


pgimalac (approved):

.gitlab/test.yml
```diff
@@ -81,6 +80,7 @@ push_to_datadog_agent:
   script:
     - source /root/.bashrc
     - set -o pipefail
+    - REF=$(echo "$CI_COMMIT_MESSAGE" | grep -oP '(?<=\[push_to_datadog_agent\]\[)[^\]]+' || echo 'main')
```
Ishirui, line 85: Oof this is quite kludgy - worth a comment explaining what this does at least.
In general would it not be better to trigger a manual job using the gitlab API (or even do everything inside github actions) and pass the ref / other params explicitely instead of relying on the auto-generated pipeline on push ?

Ishirui (approved):
Approving but really not a fan - but refactoring this would be probably quite some work

.gitlab/test.yml
Enzu83: I agree, having a comment here is relevant. I'll add one

As for the job itself, moving it entirely to a github action will be tricky because we need info such as the image tag.

I'm fine with making it manual and passing the inputs, as long as it remains simple to trigger it from a github action or elsewhere.

jinh-labs:
👋  I wanted to flag some changes in PR #1310 that may affect this PR.

- Rename `push_to_datadog_agent` job -> `push_to_datadog_agent_go_update`
- Commit message tag change `[push_to_datadog_agent]` -> `[push_to_datadog_agent_go_update]` to avoid accidental triggers on commits that merely mention the tag in their message text (more specific).
- `da-branch` tag is embedded in the commit message to pass the datadog-agent branch name reliably through squash merges; needed by the new `push_to_datadog_agent_go_update_post_merge` job to identify which datadog-agent branch to update with the final image tag after buildimages merges.

 Note: for release branch go-updates, both this PR and #1310 need to land for the full flow to work correctly.

PR #1213, opened 2026-06-19
Stop the `bazel` install test from bloating an image layer

rdesgroppes:
### What does this PR do?
Also delete any directories `bazelisk`'s `bazel` install test may produce, so nothing useless lingers in the image layer:
- on Linux, that just means handing the temporary directory through `XDG_CACHE_HOME`, **which both `bazelisk` and `bazel` honor** (resp. `$XDG_CACHE_HOME/bazelisk` and `$XDG_CACHE_HOME/bazel`),
- on Windows, `bazel` still ignores `XDG_CACHE_HOME`, but we pass it anyway should bazelbuild/bazel#27808 be addressed at some point.

### Motivation
The smoke test only has to prove that `bazelisk` can bootstrap `bazel`, whatever it downloads or unpacks is dead weight afterward.

The scripts already discard `bazelisk`'s own downloads, but `bazel` may keep an auto-extracted install base in the user's home directory, outside the directory they wipe.

Left there, that scratch would keep riding along in the Docker image layer, inflating it for no reason with about ~200MB (self-extracting executable, `A-server.jar`, `embedded_tools/`, etc.).

None: the smoke test still bootstraps `bazel` through `bazelisk` just as before.

### Additional Notes
It's really not worth keeping the auto-extracted `bazel` installation: the version it ships is almost guaranteed to be out-of-sync with any downstream branch's `.bazelversion`.

### DIFF
--- a/linux/Dockerfile
+++ b/linux/Dockerfile
@@ -318,8 +318,7 @@ ARG OSXCROSS_PATH="${DD_BUILD_INSTALL_ROOT}/osxcross" \
     RVM_PATH="${DD_BUILD_INSTALL_ROOT}/rvm"
 
 # Cache directories
-ENV BAZELISK_HOME="${XDG_CACHE_HOME}/bazelisk" \
-    DOTSLASH_CACHE="${XDG_CACHE_HOME}/dotslash" \
+ENV DOTSLASH_CACHE="${XDG_CACHE_HOME}/dotslash" \
     GOPATH="${XDG_CACHE_HOME}/go" \
     GOCACHE="${XDG_CACHE_HOME}/go/build" \
     GOMODCACHE="${XDG_CACHE_HOME}/go/pkg/mod" \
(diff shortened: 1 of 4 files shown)


linux/Dockerfile
```diff
@@ -311,8 +311,7 @@ ARG OSXCROSS_PATH="${DD_BUILD_INSTALL_ROOT}/osxcross" \
     RVM_PATH="${DD_BUILD_INSTALL_ROOT}/rvm"
 
 # Cache directories
-ENV BAZELISK_HOME="${XDG_CACHE_HOME}/bazelisk" \
```
ofek, line 321: This is intentionally defined so that usage at runtime in CI or a dev env stores data underneath our known cache root directory.

rdesgroppes: I'm with you but thought it was redundant since `bazelisk` and all Go applications leveraging `os.UserCacheDir()` honor `XDG_CACHE_HOME` - on Linux, at least:
- call site: https://github.com/bazelbuild/bazelisk/blob/5a7ff3892fdbd05672e6b8e60f139e05b2b76940/core/core.go#L359
- defintiion: https://github.com/golang/go/blob/fd6f414c65e61a51cf12c98ef473957d73f97c44/src/os/file.go#L528-L529

@ofek Do you still want me to revert it?

ofek: Yes, as I was favoring tool-specific environment variables here so that the Windows build image will eventually look the same. Can you please add a comment about that?

rdesgroppes: will do

rdesgroppes: ... done in 2nd commit

ofek (approved):
Awesome!

PR #1232, opened 2026-07-03
feat(aix): add AIX cross-compiler to Linux build image

pgimalac:
## What does this PR do?

Downloads the pre-built GCC/PowerPC→AIX cross-compiler from MASS and installs it at `/opt/aix-cross/` in the Linux x64 build image.

### Structure at `/opt/aix-cross/`

```
/opt/aix-cross/
├── compiler/          # GCC 13.4.0 targeting powerpc-ibm-aix7.3.0.0
│   ├── bin/           # powerpc-ibm-aix7.3.0.0-{gcc,g++,ld,ar,...}
│   ├── lib/, libexec/ # compiler internals + target libs (libgcc_s.a, libstdc++.a)
│   └── powerpc-ibm-aix7.3.0.0/
└── sysroot/           # AIX 7.3 headers + libraries (libc.a, crt*.o, /usr/include/...)
```

### How it's built

The toolchain was built "manually" for now for simplicity, but should be included in CI / bazel later.

### Usage

A wrapper script setting `--sysroot=/opt/aix-cross/sysroot` at invocation time makes the toolchain relocatable. With the cross-compiler on `PATH`:

```sh
export GOOS=aix GOARCH=ppc64
dda inv linter.go  # cross-linting for AIX (see datadog-agent#53082)
```


linux/Dockerfile
```diff
@@ -140,6 +140,31 @@ RUN git clone https://github.com/tpoechtrager/osxcross.git /tmp/osxcross && \
 
 RUN --mount=type=bind,src=./linux/scripts/osxcross.sh,dst=/mnt/osxcross.sh cd /tmp/osxcross && /mnt/osxcross.sh && rm -rf /tmp/osxcross
 
+FROM build_base AS aix_cross_builder
+# AIX cross-compiler (GCC targeting IBM AIX/PowerPC) downloaded from MASS.
```
aiuto, line 144: Can you add a link to the docs for how to update and publish this?

aiuto (approved):

linux/Dockerfile
```diff
@@ -140,6 +140,31 @@ RUN git clone https://github.com/tpoechtrager/osxcross.git /tmp/osxcross && \
...
+# The archive contains compiler/ (binaries + target libs) and sysroot/ (AIX headers + libs).
+# Thin wrapper scripts in bin/ call the real compiler with --sysroot so the
+# toolchain is relocatable regardless of where the sysroot was during the build.
+# Only installed on amd64: the cross-compiler itself is an x86_64 ELF binary.
+ARG TARGETARCH
+ARG AIX_CROSS_VERSION
+RUN mkdir -p /opt/aix-cross && \
+    if [ "$TARGETARCH" = "amd64" ]; then \
+    curl -fsSL "https://mass-read.us1.ddbuild.io/internal/artifact/datadog-agent/aix/cross-compiler/${AIX_CROSS_VERSION}/aix-cross-toolchain-${AIX_CROSS_VERSION}.tar.xz" | \
```
chouquette: I'd feel better if we also included a checksum/signature to ensure integrity

chouquette (approved):
integrity checks aside, LGTM

linux/Dockerfile
Ishirui: Can we use an `ADD` here ? With a checksum check like @chouquette suggested

linux/Dockerfile
```diff
@@ -140,6 +140,31 @@ RUN git clone https://github.com/tpoechtrager/osxcross.git /tmp/osxcross && \
...
+        name="$(basename "$tool")"; \
+        case "$name" in \
+            *-gcc|*-gcc-*|*-g++|*-cpp) \
+                printf '#!/bin/sh\nexec "%s" --sysroot=/opt/aix-cross/sysroot "$@"\n' "$tool" \
+                    > "/opt/aix-cross/bin/$name"; \
+                chmod +x "/opt/aix-cross/bin/$name" ;; \
+            *) ln -s "$tool" "/opt/aix-cross/bin/$name" ;; \
+        esac; \
+    done; \
+  fi
```
Ishirui: I'd really prefer to avoid having inline bash like this in the Dockerfile ; the preffered pattern is to create a quick script in `scripts` and bind-mount it at execution time so it does not get persisted in the final image.

See the `dpkg.sh` script for example (a few RUNs above)

Ishirui (changes requested):
RC because I am evil 👿 
Feel free to ping me for re-review 👍

linux/Dockerfile
pgimalac: https://datadoghq.atlassian.net/wiki/spaces/ARUN/pages/6940788504/Updating+the+AIX+cross-compiler+artifact+MASS

Ishirui (approved):

PR #1275, opened 2026-07-16
Fix trigger_tests job to use current CI_IMAGE_* buildimages variables

KevinFairise2:
## Summary
- `datadog-agent`'s `.gitlab-ci.yml` no longer defines `DATADOG_AGENT_BUILDIMAGES`/`WINBUILDIMAGES`/`ARMBUILDIMAGES` — it now uses per-image `CI_IMAGE_<TYPE>` / `CI_IMAGE_<TYPE>_SUFFIX` variables. `trigger_tests` was still setting the old, unused variables, so triggering a test pipeline against `datadog-agent` had no effect on which images it used.
- Updated `trigger_tests` to override the current variables for every image this repo always builds (linux, docker_x64/arm64, rpm_x64/arm64/armhf, btf_gen, gitlab_agent_deploy).
- Windows images (`CI_IMAGE_WIN_LTSC2022_X64`/`2025_X64`) are only overridden when the corresponding `build_windows_ltsc2022_x64`/`2025_x64` jobs actually run this pipeline (mirroring their own `rules:` in `build.yml`), so we don't point `datadog-agent` at a nonexistent test image when Windows wasn't rebuilt.

## Test plan
- [ ] Validated `.gitlab/test.yml` parses as YAML
- [ ] Manually trigger `trigger_tests` on this branch's pipeline and confirm the downstream `datadog-agent` pipeline picks up the test buildimages tags
- [ ] Trigger it again on a run that doesn't touch `windows/**` and confirm the Windows variables are left unset (falling back to `datadog-agent` defaults)


windows/README.md
```diff
@@ -1,5 +1,7 @@
 # Building on Windows
 
+<!-- touch: verify trigger_tests picks up CI_IMAGE_WIN_* when windows/** changes -->
```
rdesgroppes: leftover?

KevinFairise2: Totally, removing it, just wanted to make sure the Windows image was also working fine because this is what caused the issue first 😓

quentinus95 (approved):

PR #1281, opened 2026-07-22
Add jobs to build & publish our toolchains

chouquette:
### What does this PR do?

Extracts building the linux cross-toolchains (crosstool-ng, for x86_64, aarch64, and armhf) out of the container build jobs into a standalone CI stage, and uploads the resulting artifacts to a public S3 bucket.

This PR only adds the new toolchain-building jobs, it doesn't change how the existing build images work yet. A follow-up PR switches `linux/Dockerfile` over to consuming these prebuilt artifacts instead of building crosstool-ng in-image.

### Motivation

Decouple our toolchains from our build images. We can now update the toolchains we use (including in bazel) directly from this repo, without a manual upload, and build an armhf cross-compiler without needing the rpm-armhf build images (so we can eventually drop those for the IOT agent).

### Additional Notes

`validate_tools` failure is due to Fabric Egress Gateway being enabled for the PR, I reached out in the related incident channel to get some help, but it shouldn't block the review, only the merge :smiling_face_with_tear:

Related test PR on the agent side: https://github.com/DataDog/datadog-agent/pull/54162


.gitlab/setup.yml
```diff
@@ -32,3 +32,19 @@ validate_tools:
     - source .venv/bin/activate
     - pip install "dda==${DDA_VERSION}"
     - dda run tools check --integrity
+
+build_linux_build_base_x64:
```
alopezz: The name seems misleading, at first I thought that some of this was factoring out to also act as a base for the regular linux build image, which doesn't seem the case (and it's better that that's not the case).

Given its use, I'd probably prefer making the name clearer (specific for toolchain building). I'd also probably prefer to not have other images depend on this one (like `btf-gen`), it makes things brittle and less self-contained (relying on Gitlab to encode dependencies and making it hard to build images out of it).

I'd also keep an eye for the fact that images are not automatically pushed, see this:
https://github.com/DataDog/datadog-agent-buildimages/blob/9046a1a6df9d2cae016e063e1a008f04b3792fc9/build.sh#L41-L44

toolchains/scripts/s3-artifact-exists.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+KEY="$1"
+
+STATUS=$(curl -s -o /dev/null -w "%{http_code}" "https://dd-agent-build-artifacts.s3.amazonaws.com/${KEY}")
```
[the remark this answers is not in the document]
chouquette: that's very fair, fixed

linux/Dockerfile
```diff
@@ -83,31 +83,21 @@ FROM build_base AS texinfo_builder
...
-COPY linux/${CTNG_ARCH}/config-${CTNG_CROSS_ARCH}-unknown-gnu-linux-glibc${CROSS_GLIBC_VERSION} /tmp/crosstool-ng/crosstool-ng-${CTNG_VERSION}/.config-${CTNG_CROSS_ARCH}
-COPY linux/ctng.patch /root/ctng.patch
-RUN --mount=type=bind,src=./linux/scripts/ctng.sh,dst=/mnt/ctng.sh cd /tmp/crosstool-ng && /mnt/ctng.sh && rm -rf /tmp/crosstool-ng
+ARG NATIVE_TOOLCHAIN_KEY
+ARG CROSS_TOOLCHAIN_KEY
+ARG NATIVE_TOOLCHAIN_CHECKSUM
+ARG CROSS_TOOLCHAIN_CHECKSUM
+
+ADD --checksum=${NATIVE_TOOLCHAIN_CHECKSUM} https://dd-agent-build-artifacts.s3.amazonaws.com/${NATIVE_TOOLCHAIN_KEY} /tmp/native-toolchain.tar.xz
+ADD --checksum=${CROSS_TOOLCHAIN_CHECKSUM} https://dd-agent-build-artifacts.s3.amazonaws.com/${CROSS_TOOLCHAIN_KEY} /tmp/cross-toolchain.tar.xz
```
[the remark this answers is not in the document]
chouquette: I slightly improved the situation by simplifying the actions needed to set these variables, but as things stand, that would indeed make local builds more complicated.
I'd tend to think that it's OK since we probably only want to build locally when debugging/doing something specific, in which case executing a script prior to building is fine, but it's not optimal

toolchains/scripts/resolve-toolchain-hash.sh
```diff
@@ -0,0 +1,20 @@
...
+
+HOST_ARCH="$1"
+TARGET_ARCH="$2"
+TARGET_GLIBC_VERSION="$3"
+
+HASH=$(git log -1 --format=%H -- \
+    "toolchains/crosstool-ng/${HOST_ARCH}/config-${TARGET_ARCH}-unknown-gnu-linux-glibc${TARGET_GLIBC_VERSION}" \
+    toolchains/crosstool-ng/ctng.patch \
+    toolchains/crosstool-ng/ctng-version.env \
+    toolchains/scripts/build-crosstool-ng-toolchain.sh)
```
alopezz: Why not use a hash of the files to avoid depending on git for this?

.gitlab/build.yml
```diff
@@ -78,7 +78,19 @@
 
 build_linux_amd64:
   extends: [.build_bake, .x64]
+  before_script:
+    - !reference [.build, before_script]
+    - eval "$(./toolchains/scripts/resolve-toolchain-build-args.sh x86_64 x86_64 2.17)"
+    - eval "$(./toolchains/scripts/resolve-toolchain-build-args.sh x86_64 aarch64 2.23)"
+  needs:
+    - job: build_toolchain_x86_64_native
+      artifacts: false
```
alopezz: Why not simply use the artifacts here? Would that help us avoid the need for all the machinery around channels?

alopezz (commented):
I've got a few preliminary questions over a few parts of the design itself.

.gitlab/build.yml
chouquette: Since we need to upload to S3 in anycase (in order to be able to refer to the toolchain from datadog-agent's bazel config) it felt logical to simply always rely on the file uploaded there rather than duplicate it to gitlab

.gitlab/setup.yml
chouquette: That's fair. I initially set out to split out a base image for all or most of the existing build images, but that's not the main point for this PR, and I'm not sure it brings much in the end. I'll stick to using it only from the toolchains building jobs

toolchains/scripts/resolve-toolchain-hash.sh
chouquette: Initially this felt like a simpler approach, but manually computing the hash makes the code resilient about git commit sha changing after a rebase or similar, and this also plays better with shallow clones.
TL:DR, I switched to manual hash computation

.gitlab/build.yml
alopezz: I do see that, it's just that if that's the only reason we need to have the channel distinction it might be ultimately simpler to rely on gitlab artifacts instead. I'm a bit wary of rolling out too much ad-hoc "release engineering" for this, that's all.

chouquette: Ah, the other reason is that the bucket is configured to discard everything in `/branches` after 7 days while those in `/main` don't expire

I don't mind using gitlab artifacts between toolchain & container jobs, but with the current bucket setup, I think we still need a very basic distinction between `main` and PRs to avoid piling up toolchains

That being said, nothing is set in stone and we can definitely change course. The bucket is brand new and unused, so we can do whatever we want

alopezz: We clarified this off-line, the extra "channel" is a necessary evil if we want to test against a branch of this repo.

.gitlab/toolchains.yml
```diff
@@ -0,0 +1,79 @@
+---
+build_linux_build_base_x64:
+  extends: [.fast_build, .always, .x64]
+  stage: toolchains
+  needs: []
+  variables:
+    DOCKERFILE: linux-build-base/Dockerfile
+    IMAGE: linux-build-base
```
alopezz: A nit, but `linux-build-base` gives us very little information about the image; given that it's made to build the toolchains it might be worth calling it `toolchain-builder` or some such.

.gitlab/toolchains.yml
```diff
@@ -0,0 +1,57 @@
...
+    KUBERNETES_CPU_LIMIT: "16"
+    KUBERNETES_MEMORY_REQUEST: "16Gi"
+    KUBERNETES_MEMORY_LIMIT: "32Gi"
+
+.toolchain_build_x86_64_host:
+  extends: .toolchain_build
+  tags: ["arch:amd64"]
+  image: "registry.ddbuild.io/ci/datadog-agent-buildimages/linux-build-base${ECR_TEST_ONLY}:v${CI_PIPELINE_ID}-${CI_COMMIT_SHORT_SHA}-amd64"
+  needs:
+    - job: build_linux_build_base_x64
```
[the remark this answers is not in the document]
alopezz: This problem seems to still exist, so we probably need to gate this job to not run on those cases or make sure these are available regardless.

One other thing that we could attempt here is to actually build the docker image in the same step, so that we don't need to send it between jobs via the registry. Assuming the docker registry cache works as it should that could actually be reasonable.

toolchains/scripts/build-and-publish-toolchain.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+source ./toolchains/scripts/lib.sh
+
+TOOLCHAIN_HASH=$(resolve_toolchain_hash "${TOOLCHAIN_HOST_ARCH}" "${TOOLCHAIN_TARGET_ARCH}")
+echo "Resolved toolchain hash: ${TOOLCHAIN_HASH}"
+
+if CHANNEL=$(resolve_toolchain_channel "${TOOLCHAIN_HOST_ARCH}" "${TOOLCHAIN_TARGET_ARCH}" "${TOOLCHAIN_HASH}"); then
```
alopezz: Mostly a nit and you can leave it as is, but it's not obvious that a function called `resolve_toolchain_channel` might indicate via its return value whether the toolchain exists or not, it looks a bit overloaded to me.

toolchains/scripts/publish-toolchain.sh
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+source ./toolchains/scripts/lib.sh
+
+curl -sSf "https://awscli.amazonaws.com/awscli-exe-linux-$(uname -m).zip" -o awscliv2.zip
+unzip -q awscliv2.zip
+./aws/install
```
alopezz: This sounds like it might belong in the dockerfile.

alopezz (commented):
It's mostly looking good to me, my only real concern out of my comments is the one about the image that we depend upon not being guaranteed to be pushed under certain cases.

chouquette:
> This problem seems to still exist, so we probably need to gate this job to not run on those cases or make sure these are available regardless.

Indeed, I forgot to link this PR to the description, but instead of tagging another jobs with `.always` or similar, I removed the tag as it serves no purpose:
https://github.com/DataDog/datadog-agent-buildimages/pull/1295

toolchains/scripts/publish-toolchain.sh
chouquette: indeed, moved to the dockerfile

toolchains/scripts/build-and-publish-toolchain.sh
chouquette: I cleaned up the function to avoid returning an error code & a value. It now just returns a value.

Also split the statement to first assign the value, and then compare it, instead of shoving everything into the same one

.gitlab/toolchains.yml
chouquette: renamed the jobs to `build_toolchain_build_image_$arch` and the images to `toolchain-builder`

toolchain-builder/Dockerfile
```diff
@@ -0,0 +1,28 @@
+ARG BASE_IMAGE_REGISTRY=docker.io
+FROM ${BASE_IMAGE_REGISTRY}/ubuntu:22.04
```
hosseinsia, line 2: can we use 24.04 here? -- and preferably gbi?

chouquette:
> One other thing that we could attempt here is to actually build the docker image in the same step, so that we don't need to send it between jobs via the registry. Assuming the docker registry cache works as it should that could actually be reasonable.

That should work, but it means running the actual build as a docker in docker setup, which I'd rather avoid (I'm not sure we'd still get access to the same credentials the original pod has, and my understanding is that it's generally frowned upon when avoidable)

toolchain-builder/Dockerfile
chouquette: Currently, no for 24.04, these images are used on local developers machines, and some still run on 22.04.

When building under 24.04, the resulting binaries end up depending on some function that were introduced in the newer glibc version shipped by 24.04 and the binaries can't run on 22.04

chouquette: About using a GBI, are those publicly available or are they internal?

alopezz (approved):

alopezz:
> > One other thing that we could attempt here is to actually build the docker image in the same step, so that we don't need to send it between jobs via the registry. Assuming the docker registry cache works as it should that could actually be reasonable.
> 
> That should work, but it means running the actual build as a docker in docker setup, which I'd rather avoid (I'm not sure we'd still get access to the same credentials the original pod has, and my understanding is that it's generally frowned upon when avoidable)

You're of course right.

linux/Dockerfile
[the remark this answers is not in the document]
Ishirui: It's not ideal, but I guess we could do a bare `ADD`, no checksum and then do a checksum verification via a RUN script. If the checksum is not provided (or we find some other way to signify a local build), we could skip checksum verification.

.github/CODEOWNERS
```diff
@@ -31,6 +31,10 @@ dda.env                             @DataDog/agent-devx
 /rpm-x64/                           @DataDog/agent-devx @DataDog/agent-build
 /linux/                             @DataDog/agent-devx @DataDog/agent-build
 
+# Cross-toolchains
+/toolchains/                        @DataDog/agent-devx @DataDog/agent-build
```
Ishirui: I wonder if we should even be owning this piece, TBH I have 0 idea how crosstool-ng / the toolchains are actually used in the agent build, that seems like it's entirely you guys' prerogative 😁

.gitlab/toolchains.yml
```diff
@@ -0,0 +1,81 @@
+---
+build_toolchain_build_image_x64:
```
Ishirui, line 5: nit: I would maybe add a comment explaining in clear detail that "this is the build job for the buildimage that is used to build the toolchains that are then used in the `linux` buildimage"  - it's a mouthful, but it _is_ describing a sorta complex role 😅

toolchain-builder/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update -qy && apt-get install -y --no-install-recommends ca-certificates
+
+COPY --from=registry.ddbuild.io/images/datadog-ca-certs:standard /certs/ /usr/local/share/ca-certificates/
+RUN chmod -R o-w /usr/local/share/ca-certificates && update-ca-certificates
+
+RUN apt-get update -qy && apt-get install -y --no-install-recommends \
```
Ishirui, line 6: Why do this in two layers ? Would this fail due to missing certificates ?

Ishirui: Also, nit: add `&& rm -rf /var/lib/apt/lists/*` after every apt-get install to avoid storing useless cache

toolchain-builder/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+    unzip \
+    xz-utils \
+    ninja-build \
+    libncurses5-dev \
+    automake autoconf autopoint gettext help2man \
+    libtool libtool-bin gperf bison flex gawk pkg-config \
+    patch texinfo perl python3-dev gnupg dirmngr gpg-agent rsync
+
+ARG AWSCLI_VERSION=2.27.30
+RUN curl --retry 10 -fsSLo awscliv2.zip "https://awscli.amazonaws.com/awscli-exe-linux-$(uname -m)-${AWSCLI_VERSION}.zip" && \
```
Ishirui: nit: can we use `ADD` here instead ? And then a separate RUN step for the install

toolchains/scripts/build-crosstool-ng-toolchain.sh
```diff
@@ -0,0 +1,38 @@
...
+
+CONFIG="${CI_PROJECT_DIR}/$(toolchain_config_path "${TOOLCHAIN_HOST_ARCH}" "${TOOLCHAIN_TARGET_ARCH}")"
+TRIPLET=$(toolchain_triplet "${TOOLCHAIN_TARGET_ARCH}")
+
+WORKDIR=$(mktemp -d)
+echo "Building crosstool-ng in ${WORKDIR}"
+cd "${WORKDIR}"
+
+gpg --keyserver hkps://keyserver.ubuntu.com:443 --recv-keys 1F30EF2E
+curl -LO "https://github.com/crosstool-ng/crosstool-ng/releases/download/crosstool-ng-${CTNG_VERSION}/crosstool-ng-${CTNG_VERSION}.tar.xz"
```
Ishirui: nit: `--retry`

Ishirui (approved):
Tysm for the work on this ! I can't wait for buildimages CI to not take 1h30 and fail 2 out of 3 times 😁

.github/CODEOWNERS
chouquette: I hesitated but since most of the files are co-owned, I went with consistency, but I'll expropriate you from these files :imp:

.gitlab/toolchains.yml
chouquette: done locally

toolchains/scripts/build-crosstool-ng-toolchain.sh
chouquette: done

toolchain-builder/Dockerfile
chouquette: done, and I added a checksum verification while at it

toolchain-builder/Dockerfile
chouquette: consolidated all `apt-get install` into a single RUN and cleaned the cache

PR #1290, opened 2026-07-30
Restore the Omnibus Git cache in the Linux dev env

ofek:
### Motivation

Reapplies the variable removed in https://github.com/DataDog/datadog-agent-buildimages/pull/1282 just for our developer environments.


quentinus95 (approved):

alopezz:
Why is this needed? I'd be surprised if anyone was relying on this, and we'd rather start avoiding the omnibus git cache altogether.

ofek:
I don't expect the setting to last as the savings only exist for definitions Bazel hasn't absorbed yet. However, until its complete removal I'd prefer the functionality to be configurable as it saves about 4 minutes for me locally per build. It's quite large:

```
❯ dust -d0 $env.OMNIBUS_GIT_CACHE_DIR
376M ┌── git-cache
```

PR #1292, opened 2026-07-30
fix(dev-envs/linux): tolerate sockets when chowning the home dir

wdhif:
### What

The Linux dev-env entrypoint (`dev-envs/linux/init/entrypoint.sh`) runs `chown -R "${TARGET_USER}:" "${TARGET_HOME}"` on first start. Under `set -e` this aborts the whole entrypoint — the container exits and dev container startup fails — whenever `${TARGET_HOME}` contains a node `chown` can't operate on.

Concrete trigger: a bind-mounted `~/.ssh` that holds a **unix-domain socket** (e.g. a live ssh-agent socket under `~/.ssh/agent/`). Chowning that socket over **virtiofs** (Colima / OrbStack on macOS) fails with `ENOENT` ("No such file or directory"), so the entrypoint dies (surfaces as exit 137 to the dev container CLI). It's intermittent because the socket set is volatile.

### Fix

Chown everything under the home **except sockets and FIFOs** (which don't need it):

```sh
find "${TARGET_HOME}" ! -type s ! -type p -exec chown -h "${TARGET_USER}:" {} +
```

`-h` operates on symlinks themselves, matching `chown -R`'s default and avoiding dangling-link failures.

### Testing

Reproduced against a real unix socket over Colima virtiofs: the old `chown -R` returns `1` with the `ENOENT` error; the new `find`-based chown returns `0` on the same tree. Full `devcontainer up` (entrypoint + install-tools + deps) then completes green on macOS/Colima.

### Impact

Unblocks the Docker Desktop → Colima migration on macOS for anyone with a socket under a mounted dotfile dir. No behavior change for regular files; only sockets/FIFOs are skipped.

### DIFF
--- a/dev-envs/linux/init/entrypoint.sh
+++ b/dev-envs/linux/init/entrypoint.sh
@@ -88,8 +88,13 @@ if [[ ! -f "${startup_indicator}" ]]; then
         exit 1
     fi
 
-    # Make the target home match the runtime user before the first startup creates runtime state.
-    chown -R "${TARGET_USER}:" "${TARGET_HOME}"
+    # Make the target home match the runtime user before the first startup creates runtime
+    # state. Stay on the home's own filesystem (-xdev): bind-mounted dotfiles (e.g. ~/.ssh)
+    # live on another device, and chowning nodes there can fail for backend-specific reasons
+    # (a unix socket over virtiofs errors with ENOENT, aborting this entrypoint under
+    # `set -e`) and would needlessly rewrite host-side ownership. `-h` chowns symlinks
+    # themselves rather than dereferencing to their targets.
+    find "${TARGET_HOME}" -xdev -exec chown -h "${TARGET_USER}:" {} +
     chmod 0755 "${TARGET_HOME}"
 
     # Persist environment for SSH sessions


dev-envs/linux/init/entrypoint.sh
```diff
@@ -89,7 +89,11 @@ if [[ ! -f "${startup_indicator}" ]]; then
     fi
 
     # Make the target home match the runtime user before the first startup creates runtime state.
-    chown -R "${TARGET_USER}:" "${TARGET_HOME}"
+    # Skip sockets and FIFOs: a bind-mounted home dotfile can contain one (e.g. a live
+    # ssh-agent socket under ~/.ssh/agent), and chowning such a node fails over virtiofs
+    # (Colima/OrbStack) with ENOENT, which would abort this entrypoint under `set -e`.
+    # `-h` operates on symlinks themselves, matching `chown -R` and avoiding dangling links.
```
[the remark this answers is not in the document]
wdhif: Thanks for flagging. One correction on the mechanics: GNU `chown -R` doesn't dereference symlinks — it defaults to `-P`, so it changes the symlink itself, not the referent. Verified on coreutils 9.4: both `chown -R` and `chown -h` leave the referent untouched and only change the link's own ownership, so `-h` here preserves the existing behavior rather than changing it.

That said, the `chown -R` comparison was the confusing part, so I've reworded the comment to explain the no-dereference rationale directly.

ofek (approved):
Thanks, I might circle back on another improvement next week!

dev-envs/linux/init/entrypoint.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash -l
```
ofek, line 1: This is necessary to keep as we rely on startup files to persist changes to the environment.

ofek (changes requested):

dev-envs/linux/init/entrypoint.sh
wdhif: Reverted to `#!/bin/bash -l` — agreed the login shell's startup-file sourcing is worth keeping. (The drop was the optional hardening from your assessment; happy to leave it out.) `-xdev` chown change unchanged. Thanks!

PR #1293, opened 2026-07-31
feat: add `ddgl` to dev env

Ishirui:
### What does this PR do?

Adds DataDog/ddgl-cli to the dev envs so that our developers can use it.

### Motivation

It's a cool tool ! And it's also required for the future `/babysit-pr` skill


quentinus95 (approved):
Assuming you have double checked that changing uv venv folder for dda to mutualize it with ddgl does not bring any side effects

PR #1294, opened 2026-08-03
Use pre-built toolchains in linux build images

chouquette:
### What does this PR do?

Switches `linux/Dockerfile` to fetch the pre-built x86_64/aarch64 crosstool-ng toolchains published by #1281 from S3, instead of building crosstool-ng from source in-image.

### Motivation

Now that toolchains are built and published as standalone artifacts (#1281), the build images can just consume them, cutting the crosstool-ng build out of every image build.

Stacked on #1281.


.gitlab/build.yml
```diff
@@ -78,6 +78,15 @@
...
   extends: [.build_bake, .x64]
+  before_script:
+    - !reference [.build, before_script]
+    - eval "$(./toolchains/scripts/resolve-toolchain-build-args.sh x86_64 x86_64)"
+    - eval "$(./toolchains/scripts/resolve-toolchain-build-args.sh x86_64 aarch64)"
+  needs:
+    - job: build_toolchain_x86_64_native
+      artifacts: false
+    - job: build_toolchain_x86_64_cross_aarch64
+      artifacts: false
```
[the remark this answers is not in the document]
chouquette: That's valid, but on the other hand, this repo doesn't have tags, so instead of adding another tag to new jobs, I'd rather remove it entirely since it serves no purpose. I'll open a follow up PR

linux/Dockerfile
```diff
@@ -83,31 +83,21 @@ FROM build_base AS texinfo_builder
...
-RUN --mount=type=bind,src=./linux/scripts/ctng.sh,dst=/mnt/ctng.sh cd /tmp/crosstool-ng && /mnt/ctng.sh && rm -rf /tmp/crosstool-ng
+ARG NATIVE_TOOLCHAIN_KEY
+ARG CROSS_TOOLCHAIN_KEY
+ARG NATIVE_TOOLCHAIN_CHECKSUM
+ARG CROSS_TOOLCHAIN_CHECKSUM
+
+ADD --checksum=${NATIVE_TOOLCHAIN_CHECKSUM} https://dd-agent-build-artifacts.s3.amazonaws.com/${NATIVE_TOOLCHAIN_KEY} /tmp/native-toolchain.tar.xz
+ADD --checksum=${CROSS_TOOLCHAIN_CHECKSUM} https://dd-agent-build-artifacts.s3.amazonaws.com/${CROSS_TOOLCHAIN_KEY} /tmp/cross-toolchain.tar.xz
+
+RUN mkdir -p /opt/toolchains/${CTNG_ARCH} /opt/toolchains/${CTNG_CROSS_ARCH} \
```
aiuto: It sort of feels like this should be 2 RUN commands, one for each ADD.

toolchain-builder/Dockerfile
aiuto: Do we really need both curl and wget?

aiuto (approved):

toolchain-builder/Dockerfile
chouquette: We don't, I changed ctng configs so that it also uses curl instead of the default wget (although that change was made in the base PR, #1281)

docker-bake.hcl
```diff
@@ -154,6 +154,43 @@ variable "misc_args_arm64" {
   }
 }
 
+// Resolved per-pipeline in CI (see .gitlab/build.yml) from the recipe files under
+// toolchains/, and used by crosstool_ng_builder to fetch pre-built toolchains from
+// the dd-agent-build-artifacts S3 bucket. To build linux-amd64/linux-arm64 locally,
+// eval toolchains/scripts/resolve-toolchain-build-args.sh first for each of the
+// native and cross toolchains, e.g.:
+//   eval "$(./toolchains/scripts/resolve-toolchain-build-args.sh x86_64 x86_64)"
+//   eval "$(./toolchains/scripts/resolve-toolchain-build-args.sh x86_64 aarch64)"
```
Ishirui: Hmmm that's not great - I'd really like to avoid regressing from `docker buildx bake` to `<some script>;docker buildx bake` for local builds. It's not so much that this one extra step is a big deal, it's that if we open the door to this kind of stuff then it's a slippery slope.

Can we instead have that script invocation _inside_ the Dockerfile, maybe gated behind a stage that only runs for local builds ?

linux/Dockerfile
Ishirui: nit: Better to use `--mount=type=tmpfs` [docs](https://docs.docker.com/reference/dockerfile/#run---mounttypetmpfs) for this sort of stuff, it's higher performance than writing into /tmp IIRC + no need to cleanup

Ishirui (changes requested):
RC because of the local-build regression, sorry :/ 
Let's discuss alternatives ?

docker-bake.hcl
aiuto: Maybe I'm not understanding this completely.

ISTM that we would never run a resolve toolchains script while building the dev image.
- the scripts for build and publish toolchains is used by the toolchains publish
- the script to create a build image must use published toolchain

Why would a local build have to use anything except a released (or release candidate) toolchain?

chouquette: > we would never run a resolve toolchains script while building the dev image.
I'd say there no good reason to prevent locally building the build images, and in order for them to be functional, they need toolchains, so they need to try and fetch one.

> the script to create a build image must use published toolchain

Yes, a local toolchain build AND a local build image is currently out of scope, but it the need arise, we could allow a local file to be used instead I suppose.

> Why would a local build have to use anything except a released (or release candidate) toolchain

As we migrate more and more code to bazel, we won't need the build images to debug issues, but until that happens, I'd argue that we need to be able to generate a local build image to reproduce build issue, and I don't see a good reason to prevent a WIP toolchain from being used in a local build

TL:DR, as long as the generated build image isn't pushed to a registry (which I don't think we can) developers should be free to use any toolchain, stable or not into their local build images

linux/Dockerfile
chouquette: done

chouquette:
@Ishirui pushed a new version which allows local build easily, resolving & fetching the toolchains directly from the dockerfile, let me know if that works better for you!

toolchains/scripts/fetch-toolchain.sh
```diff
@@ -0,0 +1,29 @@
...
+CHECKSUM=$(curl --retry 10 -sSf "${S3_BASE_URL}/${KEY}.sha256")
+CHECKSUM="${CHECKSUM#sha256:}"
+
+TARBALL=$(mktemp)
+curl --retry 10 -sSf "${S3_BASE_URL}/${KEY}" -o "${TARBALL}"
+echo "${CHECKSUM}  ${TARBALL}" | sha256sum -c -
+
+mkdir -p "${DEST}"
+tar xf "${TARBALL}" -C "${DEST}"
+rm "${TARBALL}"
```
Ishirui, line 29: nit: usually it's better to avoid doing shell-based `curl` and `sha256sum`, using [ADD](https://docs.docker.com/reference/dockerfile/#add) instead, especially for heavy tarballs - this is because ADD behaves a lot better w.r.t. docker build cache than `RUN ...`, and also because all ADD downloads start executing _immediately_ after the build starts instead of waiting on previous layers like a `RUN curl ...` does.

Here this would mean you would split the script to only fetch the required channel, key, hash etc. and then do the actual download inside the Dockerfile using ADD. But this makes the script less portable, so if we ever use it directly or in another Dockerfile that's worse. I'll let you decide on this one 👍

linux/Dockerfile
```diff
@@ -83,31 +84,14 @@ FROM build_base AS texinfo_builder
 ADD https://gnu.googlesource.com/texinfo.git#60d3edc4b74b4e1e5ef55e53de394d3b65506c47 /tmp/texinfo
 RUN --mount=type=bind,src=./linux/scripts/texinfo.sh,dst=/mnt/texinfo.sh cd /tmp/texinfo && /mnt/texinfo.sh
```
Ishirui: I don't have the historical context on this, so I am 100% unsure but: does removing the crosstool-ng build also mean we could remove this ? I don't see a good reason why we need this specific bugfix version in the buildimage, my best guess is that it was somehow related to the toolchain build.

If we could remove this stage in favor of a simple `apt install texinfo` that would be great

Ishirui (approved):
🚀

toolchains/scripts/fetch-toolchain.sh
chouquette: AFAIU this would require some heavier logic for local builds, we couldn't just `docker build` and would need a script or a Makefile to compute the link to the toolchains and pass those to `ADD`, I'm not sure it's worth it to be honest

linux/Dockerfile
chouquette: good point, there are probably quite a few leftovers. I'm not sure we even need texinfo anymore

PR #1302, opened 2026-08-12
[SINT-5606] Install Datadog CA Certs in the agent-deploy Image

cedricvanrompay-datadog:
Required to be able to reach out to internal services via Fabric DNS (needed in https://github.com/DataDog/agent-release-management/pull/478)

Note that another solution would be to base the image on a Golden Base Image (GBI) such as `registry.ddbuild.io/images/base/gbi-ubuntu_2604:release`, however there is a larger risk of regression due to the change of OS version (from Ubuntu 18 to Ubuntu 26).

### DIFF
--- a/.gitlab/setup.yml
+++ b/.gitlab/setup.yml
@@ -3,7 +3,7 @@ lint_dockerfiles:
   tags: ["arch:amd64"]
   stage: setup
   needs: []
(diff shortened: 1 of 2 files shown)


cedricvanrompay-datadog:
The failure of job `lint_dockerfiles` does not come from this PR, the job is failing on main as well.

arnaudpy (approved):

chouetz (approved):

PR #1315, opened 2026-08-20
Exclude OTEL_* environment variable from workspace image

KevinFairise2:
### What does this PR do?

Exclude OTEL_* env var from the env snapshot. Because they are injected by buildkit and we 

### Motivation

When BuildKit tracing is enabled, it injects OpenTelemetry vars into each build step, including:

OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=unix:///dev/otel-grpc.sock
OTEL_EXPORTER_OTLP_TRACES_PROTOCOL=grpc
OTEL_TRACES_EXPORTER=otlp
OTEL_TRACE_PARENT / TRACEPARENT
/dev/otel-grpc.sock is BuildKit’s in-container gRPC socket for nested traces, not something defined in the Dockerfiles.


chouetz (approved):
STGM and you should review codex's PR

